PR #59, opened 2019-02-28
First test and Framework

pritampatil4:


example/src/androidTest/java/net/optile/payment/HomePageTests.java
```diff
@@ -0,0 +1,38 @@
...
+
+
+
+
+@RunWith(AndroidJUnit4.class)
+@LargeTest
+public class HomePageTests {
+
+    @Rule
+    public ActivityTestRule<CheckoutActivity> mActivityRule = new ActivityTestRule<>(
```
siebeprojects: Please name it activityRule and not mActivityRule, we do not use this style for naming.

example/src/androidTest/java/net/optile/payment/PaymentInstrumentedTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+import android.content.Context;
+import android.support.test.InstrumentationRegistry;
+import android.support.test.runner.AndroidJUnit4;
+
+/**
+ * Instrumented test, which will execute on an Android device.
+ *
+ * @see <a href="http://d.android.com/tools/testing">Testing documentation</a>
+ */
```
siebeprojects: We can remove this class completely as it will be replaced by our own test classes.

example/src/androidTest/java/net/optile/payment/HomePageTests.java
```diff
@@ -0,0 +1,38 @@
...
+import static android.support.test.espresso.matcher.ViewMatchers.isDisplayed;
+import static android.support.test.espresso.matcher.ViewMatchers.withId;
+import static android.support.test.espresso.matcher.ViewMatchers.withText;
+
+
+
+
+@RunWith(AndroidJUnit4.class)
+@LargeTest
+public class HomePageTests {
```
siebeprojects: Name it CheckoutPageTests, since the class is also called CheckoutActivity

example/src/androidTest/java/net/optile/payment/PaymentInstrumentedTest.java
pritampatil4: Done.

PR #71, opened 2019-04-24
Pcx 182 Support finalizing charge of PresetAccount

siebeprojects:
Once a PresetAccount has been set in a list, it is now possible to perform a charge request on the PresetAccount using the Android SDK.
The PaymentUI has a new method called chargePresetAccount, this will show a progress screen while the Android SDK is sending the charge operation to the Payment API.


payment/src/main/java/net/optile/payment/form/Operation.java
```diff
@@ -73,6 +73,37 @@ public void putValue(String name, Object value) throws PaymentException {
...
+     * If the type cannot be determined from the URl then null will be returned.
+     *
+     * @return the type of the operation or null if it cannot be determined.
+     */
+    public String getType() {
+        String path = this.url.getPath();
+
+        if (path.endsWith("preset")) {
+            return PRESET;
+        } else if (path.endsWith("charge")) {
```
darklynx: for some payment methods `path` may end with `activate`

payment/src/main/java/net/optile/payment/form/Operation.java
```diff
@@ -73,6 +73,37 @@ public void putValue(String name, Object value) throws PaymentException {
...
+    }
+
+    /**
+     * Check if the type of this operation matches the given type.
+     *
+     * @param type to match with the type of this operation.
+     * @return true when the types matches, false otherwise.
+     */
+    public boolean isType(String type) {
+        return type != null && type.equals(getType());
```
darklynx:
```
return Objects.equals(type, getType());
```

siebeprojects: Will replace it with Objects.equals.

payment/src/main/java/net/optile/payment/form/Operation.java
siebeprojects: I will add the activate value as well.

payment/src/main/java/net/optile/payment/network/ListConnection.java
```diff
@@ -39,6 +41,21 @@
  */
 public final class ListConnection extends BaseConnection {
 
+    private final static Object cacheLock = new Object();
+    private final static Map<String, LanguageFile> languageCache = new HashMap<>();
+
+    private static void cacheLanguageFile(String key, LanguageFile file) {
+        synchronized (cacheLock) {
```
darklynx: Use of `Collections.synchronizedMap(...)` or even `ConcurrentHashMap` could simplify it if Android Java has those.

payment/src/main/java/net/optile/payment/ui/page/PresetAccountPresenter.java
```diff
@@ -0,0 +1,349 @@
...
+
+        dialog.setListener(new ThemedDialogListener() {
+            @Override
+            public void onButtonClicked(ThemedDialogFragment dialog, int which) {
+                switch (which) {
+                    case ThemedDialogFragment.BUTTON_NEUTRAL:
+                        view.closePage();
+                        break;
+                    case ThemedDialogFragment.BUTTON_POSITIVE:
+                        postOperation(operation);
```
darklynx: interesting syntax, I would always put a break here, just in case you get another `case` in the future...

PR #82, opened 2019-06-19
Pcx 313 smartswitch

siebeprojects:
Remove smart switch logic from the NetworkCard and PaymentNetwork classes.
This makes it easier to re-use the PaymentNetwork class for GooglePay activation.

### DIFF
--- a/payment/src/main/java/net/optile/payment/ui/model/SmartSwitch.java
+++ b/payment/src/main/java/net/optile/payment/ui/model/SmartSwitch.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright (c) 2019 optile GmbH
+ * https://www.optile.net
+ *
+ * This file is open source and available under the MIT license.
+ * See the LICENSE file for more information.
+ */
+
+package net.optile.payment.ui.model;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+
+import android.text.TextUtils;
+
+/**
+ * Class for storing and selecting payment networks based on smart switch logic
+ */
+public final class SmartSwitch {
+
+    private final HashMap<String, String> smartMapping;
+    private final List<PaymentNetwork> networks;
+    private final List<PaymentNetwork> smartSelected;
+    private final List<PaymentNetwork> smartBuffer;
+
+    public SmartSwitch(List<PaymentNetwork> networks) {
+        this.networks = networks;
+        this.smartMapping = new HashMap<>(); 
+        this.smartSelected = new ArrayList<>();
+        this.smartBuffer = new ArrayList<>();
+    }
+
+    public void addSelectionRegex(String code, String regex) {
+        smartMapping.put(code, regex);
+    }
+    
+    /**
+     * Get the list of smart selected PaymentNetworks.
+     *
+     * @return the list of smart selected PaymentNetworks.
+     */
+    public List<PaymentNetwork> getAllSelected() {
+        return smartSelected;
+    }
+
+    /**
+     * Get the first smart selected PaymentNetwork, may return null if none are selected.
+     *
+     * @return the first smart selected PaymentNetwork.
+     */
+    public PaymentNetwork getFirstSelected() {
+        return smartSelected.size() > 0 ? smartSelected.get(0) : null;
+    }
+    
+    /**
+     * Are any PaymentNetworks selected based on smart switch.
+     *
+     * @return true when there are selected payment networks, false otherwise
+     */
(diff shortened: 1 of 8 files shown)


payment/src/main/java/net/optile/payment/ui/model/SmartSwitch.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.List;
+
+import android.text.TextUtils;
+
+/**
+ * Class for storing and selecting payment networks based on smart switch logic
+ */
+public final class SmartSwitch {
+
+    private final HashMap<String, String> smartMapping;
```
hjhill: You should probably use `Map` here - that the instance is a `HashMap` is an implementation detail.

payment/src/main/java/net/optile/payment/ui/model/SmartSwitch.java
```diff
@@ -0,0 +1,107 @@
...
+        return smartSelected;
+    }
+
+    /**
+     * Get the first smart selected PaymentNetwork, may return null if none are selected.
+     *
+     * @return the first smart selected PaymentNetwork.
+     */
+    public PaymentNetwork getFirstSelected() {
+        return smartSelected.size() > 0 ? smartSelected.get(0) : null;
```
hjhill: Minor: better use `isEmpty` (might make a difference with some list implementations)

payment/src/main/java/net/optile/payment/ui/model/SmartSwitch.java
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * Check if the PaymentNetwork is smart selected, it is smart selected when the provided number input matches
+     * the regex of this PaymentMethod in the groups settings file. A PaymentMethod is always smart selected when
+     * there is only one PaymentMethod in the NetworkCard.
+     *
+     * @param network to check if smart selection
+     * @return true when smart selected, false otherwise
+     */
+    public boolean isSelected(PaymentNetwork network) {
+        if (networks.size() == 1 && networks.get(0) == network) {
```
hjhill: Almost surely an unnecessary optimization.

payment/src/main/java/net/optile/payment/ui/model/SmartSwitch.java
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * Validate the PaymentMethods for smart selection given the new text
+     *
+     * @param text to validate for smart selection
+     * @return true when the smart selection of payment methods has changed, false otherwise
+     */
+    public boolean validate(String text) {
+        smartBuffer.clear();
+        String regex;
```
hjhill: I would prefer defining the variable where it is set - less clutter.

payment/src/main/java/net/optile/payment/ui/model/SmartSwitch.java
```diff
@@ -0,0 +1,107 @@
...
+
+        if (text != null) {
+            for (PaymentNetwork network : networks) {
+                regex = smartMapping.get(network.getCode());
+
+                if (!TextUtils.isEmpty(regex) && text.matches(regex)) {
```
hjhill: Would pre-compiling the regex to a `Pattern` be better here? (for desktop Java it definitely would, but not sure about Android)

payment/src/main/java/net/optile/payment/ui/model/SmartSwitch.java
hjhill: Sorry, mixed up the two lists in my mind...

payment/src/main/java/net/optile/payment/ui/model/SmartSwitch.java
siebeprojects: Changed to Map

payment/src/main/java/net/optile/payment/ui/model/SmartSwitch.java
siebeprojects: Using isEmpty now

payment/src/main/java/net/optile/payment/ui/model/SmartSwitch.java
siebeprojects: Done

payment/src/main/java/net/optile/payment/ui/model/SmartSwitch.java
siebeprojects: Using now Pattern instead of String regex

PR #85, opened 2019-07-30
Pcx 378 - Add Process Payment screen for handling charge requests

siebeprojects:


payment/src/main/java/net/optile/payment/form/Operation.java
```diff
@@ -14,32 +14,78 @@
...
+            return new Operation(in);
+        }
+
+        public Operation[] newArray(int size) {
+            return new Operation[size];
+        }
+    };
+    private URL url;
+    private JSONObject form;
+    private JSONObject account;
```
hjhill: Why are those not `final` anymore?

payment/src/main/java/net/optile/payment/network/BaseConnection.java
```diff
@@ -284,10 +283,10 @@ PaymentException createPaymentException(final String source, final String errorT
         } catch (IOException | JsonParseException e) {
             // Ignore the exceptions since the ErrorInfo is an optional field
             // and it is more important to not loose the status error code
```
hjhill: Not changed, but still typo: `lose`

payment/src/main/java/net/optile/payment/ui/PaymentUI.java
```diff
@@ -58,6 +59,19 @@ public static PaymentUI getInstance() {
         return InstanceHolder.INSTANCE;
     }
 
+    /**
+     * Get the PaymentResult from the data intent.
+     *
+     * @param data intent to obtain the PaymentResult form
+     * @return PaymentResult or null if not stored in the intent
+     */
+    public final static PaymentResult getPaymentResult(Intent data) {
```
hjhill: Doesn't this method belong into `PaymentResult`?

payment/src/main/java/net/optile/payment/ui/model/AccountCard.java
```diff
@@ -108,11 +121,6 @@ public boolean onTextInputChanged(String type, String text) {
         return false;
     }
 
-    public URL getLink(String name) {
```
hjhill: FYI: Avoid moving code around to reduce the number of changed lines and make people more inclined to review your PRs.

payment/src/main/java/net/optile/payment/ui/page/BasePaymentActivity.java
```diff
@@ -26,6 +26,8 @@
  */
 abstract class BasePaymentActivity extends AppCompatActivity {
 
+    public final static String EXTRA_OPERATION = "operation";
```
hjhill: Will this be used by other child classes of `BasePaymentActivity`?

payment/src/main/java/net/optile/payment/ui/theme/ProcessParameters.java
```diff
@@ -0,0 +1,55 @@
...
+package net.optile.payment.ui.theme;
+
+import android.support.annotation.StyleRes;
+import net.optile.payment.R;
+
+/**
+ * Class for holding the PaymentPage theme parameters.
+ * These parameters may be used to theme the PaymentPage UI elements and text appearances.
+ */
+public final class ProcessParameters {
```
hjhill: FYI: Maybe you should think about using Lombok.

payment/src/main/java/net/optile/payment/ui/page/PaymentListPresenter.java
```diff
@@ -154,6 +148,53 @@ public void onPaymentSessionError(Throwable cause) {
...
+        if (result.requestCode != PROCESSPAYMENT_REQUEST_CODE) {
+            showPaymentSession(this.session);
+            return;
+        }
+        switch (result.resultCode) {
+            case PaymentUI.RESULT_CODE_OK:
+                view.passOnPaymentResult(PaymentUI.RESULT_CODE_OK, result.paymentResult);
+                break;
+            case PaymentUI.RESULT_CODE_ERROR:
+                view.passOnPaymentResult(PaymentUI.RESULT_CODE_ERROR, result.paymentResult);
```
hjhill: Couldn't you handle both the previous `case` and this one with the following code?

```
case PaymentUI.RESULT_CODE_OK:
case PaymentUI.RESULT_CODE_ERROR:
    view.passOnPaymentResult(result.resultCode, result.paymentResult);
    break;
```

payment/src/main/java/net/optile/payment/ui/page/ProcessPaymentActivity.java
```diff
@@ -0,0 +1,199 @@
...
+    private Operation operation;
+
+    /**
+     * Create the start intent for this ProcessPaymentActivity
+     *
+     * @param context Context to create the intent
+     * @return newly created start intent
+     */
+    public static Intent createStartIntent(Context context, Operation operation) {
+        if (context == null) {
```
hjhill: FYI: `Objects.requireNonNull` might come in handy here...

payment/src/main/java/net/optile/payment/form/Operation.java
siebeprojects: Changed them back to final, they should be final.

payment/src/main/java/net/optile/payment/network/BaseConnection.java
siebeprojects: will fix it.

payment/src/main/java/net/optile/payment/ui/PaymentUI.java
siebeprojects: yes, will move it to PaymentResult similar as to how Google / Android is doing it.

payment/src/main/java/net/optile/payment/ui/model/AccountCard.java
siebeprojects: Yes, will try either putting the code cleanup in a different commit or PR.

payment/src/main/java/net/optile/payment/ui/page/BasePaymentActivity.java
siebeprojects: Moved this to ProcessPaymentActivity since this is for now the only activity using it.

payment/src/main/java/net/optile/payment/ui/theme/ProcessParameters.java
siebeprojects: yes that would be an option, but we have to keep the external libraries to an absolute minimum in the SDKs.

payment/src/main/java/net/optile/payment/ui/page/PaymentListPresenter.java
siebeprojects: Done

payment/src/main/java/net/optile/payment/ui/page/ProcessPaymentActivity.java
siebeprojects: Yes, I agree. I would then like to change it in all classes so it remains consistent. I rather not like to change it in this PR.

payment/src/main/java/net/optile/payment/ui/model/AccountCard.java
```diff
@@ -31,6 +31,14 @@ public AccountCard(AccountRegistration account, PaymentNetwork network) {
         this.network = network;
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean containsLink(String name, URL url) {
+        return url.equals(getLink(name));
```
darklynx: Comparing 2 URLs using `equals` method sometimes can return `false` even for 2 identical URLs, I'm not sure if Android Java is affected, but we have experienced that in out functional tests on Linux.

siebeprojects: Changed this to compare the string value of both URLs with each other.

PR #86, opened 2019-08-19
Pcx 379

siebeprojects:
Two main changes are in this PR. 
1. Renaming the ProcessPayment to ChargePayment, this is because the ChargePaymentActivity only takes care of the final Charge operation. All other operations like update, preset, etc are handled by the PaymentListActivity. 
2. Move the payment logic for VISA and other basic payment methods to a BasicNetworkService. This NetworkService implementation will take care of performing the operation of the selected network.


payment/src/main/java/net/optile/payment/ui/service/NetworkServiceLookup.java
```diff
@@ -0,0 +1,66 @@
...
+    /** 
+     * Is the ApplicableNetwork supported by any of the NetworkServices provided in this Android SDK.
+     * 
+     * @param network check if this network is supported
+     * 
+     * @return true when supported, false otherwise
+     */
+    public static boolean isSupported(ApplicableNetwork network) {
+        NetworkServiceFactory factory = getFactory(network.getCode());
+        return factory != null && factory.isSupported(network);
```
hjhill: Will the `isSupported` ever return `false` here? If so, did the call to `getFactory` find the "wrong" `NetworkService` (i.e., could it find a better one by checking the other factories)?

payment/src/main/java/net/optile/payment/ui/service/NetworkServiceLookup.java
```diff
@@ -0,0 +1,66 @@
...
+
+import net.optile.payment.model.ApplicableNetwork;
+import net.optile.payment.ui.service.basic.BasicNetworkFactory;
+
+/**
+ * Class for looking up a NetworkService given the ApplicableNetwork.
+ */
+public class NetworkServiceLookup {
+
+    private static volatile List<NetworkServiceFactory> factories = new ArrayList<>();
```
hjhill: `volatile` is not necessary here, `final` might be a good idea.

payment/src/main/java/net/optile/payment/ui/service/NetworkServiceLookup.java
```diff
@@ -0,0 +1,66 @@
...
+        for (NetworkServiceFactory factory : factories) {
+            if (factory.isSupported(code)) {
+                return factory;
+            }
+        }
+        return null;
+    }
+    
+    private static void initFactories() {
+        synchronized (factories) {
```
hjhill: Synchronization is a good idea, but must be done consistently (i.e., when writing *and* reading). Other option: The `CopyOnWriteArrayList` makes synchronization unnecessary when writes are rare (or, as might be the case here, only done in an initialization phase)

BTW: At the moment there's only this `BasicNetworkFactory`, but you planned this as the point of extension, correct?

payment/src/main/java/net/optile/payment/ui/service/basic/BasicNetworkFactory.java
```diff
@@ -0,0 +1,65 @@
...
+import net.optile.payment.core.PaymentNetworkCodes;
+import net.optile.payment.model.ApplicableNetwork;
+import net.optile.payment.ui.service.NetworkService;
+import android.text.TextUtils;
+import net.optile.payment.ui.service.NetworkServiceFactory;
+
+/**
+ * Specific implementation for basic networks like i.e. Visa, mastercard and sepa.
+ */
+public final class BasicNetworkFactory implements NetworkServiceFactory {
```
hjhill: Minor: I would call it `BasicNetworkServiceFactory` (more consistent IMO).

payment/src/main/java/net/optile/payment/ui/page/ChargePaymentPresenter.java
```diff
@@ -0,0 +1,312 @@
...
+        sessionService = new PaymentSessionService();
+        sessionService.setListener(this);
+    }
+
+    void onStart(Operation operation) {
+        this.operation = operation;
+        this.listUrl = PaymentUI.getInstance().getListUrl();
+        
+        if (activityResult != null) {
+            handleActivityResult(activityResult);
```
hjhill: FYI: A more logical ordering of private methods would make reviewing easier.

payment/src/main/java/net/optile/payment/form/Operation.java
```diff
@@ -163,4 +164,31 @@ public String toJson() throws JSONException {
...
         return url;
     }
+
+    /** 
+     * Create a new empty Operation using the URL from the provided operation and use the newType.
+     *
+     * @param operation the source operation to use
+     * @param newType the type of the new Operation
+     */
+    public final static Operation createOperationFrom(Operation operation, String newType) throws PaymentException {
```
hjhill: Is this some sort of `cloneWithChangedType` operation? In any case, I'd argue that `Operation` and possibly also `From` are redundant in this name...

payment/src/main/java/net/optile/payment/form/Operation.java
```diff
@@ -163,4 +164,31 @@ public String toJson() throws JSONException {
...
+        String curType = operation.getType();
+
+        if (curType == null) {
+            throw new PaymentException("Could not determine type from operation URL");
+        }
+        String src = operation.getURL().toString();
+        String from = curType.toLowerCase();
+        String to = newType.toLowerCase();
+        int lastIndex = src.lastIndexOf(from);
+        String tail = src.substring(lastIndex).replaceFirst(from, to);
```
hjhill: `replaceFirst` searches for `from` again - this is unnecessary and may confuse readers. Better create the new string as
`src.substring(0, lastIndex) + to + src.substring(lastIndex + from.length())`.

payment/src/main/java/net/optile/payment/ui/PaymentResult.java
```diff
@@ -41,6 +43,26 @@ public PaymentResult createFromParcel(Parcel in) {
...
     private PaymentError error;
 
+    /** 
+     * Construct a new PaymentResult given the cause, the cause can either be a PaymentException or any other exception
+     * 
+     * @param cause of the error
+     */
+    public final static PaymentResult fromThrowable(Throwable cause) {
+        if (cause instanceof PaymentException) {
+            PaymentException pe = (PaymentException) cause;
```
hjhill: Would it make sense to extract the following lines into another `static constructor` (possible name: `fromPaymentException`)?

payment/src/main/java/net/optile/payment/ui/page/BasePaymentActivity.java
```diff
@@ -150,7 +150,7 @@ void setUserClosedPageResult() {
      * @param resultCode of the ActivityResult
      * @param result to be added as extra to the intent
      */
-    void setActivityResult(int resultCode, PaymentResult result) {
+    void setResultIntent(int resultCode, PaymentResult result) {
```
hjhill: I'm not sure I like this name change... let's discuss.

payment/src/main/java/net/optile/payment/ui/page/ChargePaymentPresenter.java
```diff
@@ -180,39 +174,39 @@ public void showProgress() {
      * {@inheritDoc}
      */
     @Override
-    public void onPreparePaymentResult(int resultCode, PaymentResult paymentResult) {
+    public void onPreparePaymentResult(int resultCode, PaymentResult result) {
     }
```
hjhill: I've seen this empty implementation a few times already - would an empty default implementation be sensible?

payment/src/main/java/net/optile/payment/ui/service/PaymentSessionService.java
```diff
@@ -285,6 +285,9 @@ private LanguageFile loadPaymentPageLanguageFile(Map<String, PaymentNetwork> net
     }
 
     private boolean isSupported(ApplicableNetwork network) {
+        
+
+
```
hjhill: ???

payment/src/main/java/net/optile/payment/ui/service/NetworkServiceLookup.java
siebeprojects: There will only be one NetworkService per network type. 
It may be so that this NetworkService only supports a subset of ApplicableNetwork capabilities like only non-redirect networks for VISA.

payment/src/main/java/net/optile/payment/ui/page/BasePaymentActivity.java
siebeprojects: Will change this in a later PR when we refactor by using the BaseView and BasePresenter.

payment/src/main/java/net/optile/payment/ui/page/ChargePaymentPresenter.java
siebeprojects: This will be fixed when we have a BasePresenter for Payment Views.

payment/src/main/java/net/optile/payment/ui/service/PaymentSessionService.java
siebeprojects: Correct, this method is now obsolete since it is implemented in the NetworkServiceFactory.

payment/src/main/java/net/optile/payment/ui/PaymentResult.java
siebeprojects: Added this to this PR.

payment/src/main/java/net/optile/payment/ui/service/NetworkServiceLookup.java
siebeprojects: Removed volatile and replaced with final

payment/src/main/java/net/optile/payment/ui/service/NetworkServiceLookup.java
siebeprojects: Replaced it with the CopyOnWriteArrayList.
For now, we have one basic NetworkService but GooglePay will be added as a second one.

payment/src/main/java/net/optile/payment/ui/page/ChargePaymentPresenter.java
siebeprojects: Will take care of it, I normally use Android Studio to logically reorder code.

payment/src/main/java/net/optile/payment/form/Operation.java
siebeprojects: Changed it to create and added a unitTest for testing this create method.

payment/src/main/java/net/optile/payment/ui/PaymentResult.java
siebeprojects: Added a second method called fromPaymentException.

payment/src/main/java/net/optile/payment/ui/service/basic/BasicNetworkFactory.java
siebeprojects: Changed it to BasicNetworkServiceFactory which is better.

payment/src/main/java/net/optile/payment/form/Operation.java
siebeprojects: Added this logic to create a new Operation with a different type from another one.

PR #93, opened 2019-09-17
PCX-388 Localization fallback

siebeprojects:
Replace individual language files stored in the Payment Network and Session objects with a global Localization class.
This allows the Android SDK to provide a fallback mechanism if the language file of a network does not contain the translation.


payment/src/main/java/net/optile/payment/ui/service/PaymentSessionService.java
```diff
@@ -99,45 +109,85 @@ public void loadPaymentSession(final String listUrl, final Context context) {
...
 
-        return new PaymentSession(listResult, presetCard, accountCards, networkCards, validator, lang);
+    private void loadLocalization(Context context, Map<String, PaymentNetwork> networks) throws PaymentException {
+        Localization loc = Localization.getInstance();
+        URL langUrl = null;
+        String code = null;
+        for (PaymentNetwork network : networks.values()) {
+            langUrl = network.getLink("lang");
+            if (langUrl == null) {
+                throw createPaymentException("Missing 'lang' link in PaymentNetwork", null);
```
hjhill: Would it make sense to include the network code here (better debugging)?

payment/src/main/java/net/optile/payment/ui/service/PaymentSessionService.java
```diff
@@ -99,45 +109,85 @@ public void loadPaymentSession(final String listUrl, final Context context) {
...
-
         PresetCard presetCard = createPresetCard(listResult, networks);
         List<AccountCard> accountCards = createAccountCards(listResult, networks);
         List<NetworkCard> networkCards = createNetworkCards(networks, groups);
+        return new PaymentSession(listResult, presetCard, accountCards, networkCards, validator);
+    }
 
-        return new PaymentSession(listResult, presetCard, accountCards, networkCards, validator, lang);
+    private void loadLocalization(Context context, Map<String, PaymentNetwork> networks) throws PaymentException {
+        Localization loc = Localization.getInstance();
```
hjhill: As @siebeprojects and me found out when talking about this, the localization cache should be cleared when the `listUrl` changes.

payment/src/main/java/net/optile/payment/core/Localization.java
```diff
@@ -0,0 +1,251 @@
...
+import android.text.TextUtils;
+import net.optile.payment.R;
+import net.optile.payment.model.Interaction;
+
+/**
+ * Class holding the localization files and local translations.
+ * If the translation is not available in a localization file with the given name, it will try to obtain the translation from the shared file.
+ * If the shared file is not set or does not contain the translation, then it will try to obtain the translation from the local translations map.
+ */
+public final class Localization {
```
hjhill: This class definitely does too much:
- handling localization keys
- handling local translations
- handling a shared translation "file" (loaded from a URL)
- handling translation "files" (also loaded from a URL)

payment/src/main/java/net/optile/payment/core/Localization.java
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    /**
+     * Helper method to check if a translation exist for the account hint.
+     *
+     * @param fileName the name of the Properties file
+     * @param account name of the account
+     * @return true if it has a translation for the account hint, false otherwise
+     */
+    public static boolean hasAccountHint(String fileName, String account) {
```
hjhill: Do you really need a separate method for that? If the result is `true`, you probably still need the translation somewhere...

payment/src/main/java/net/optile/payment/ui/service/PaymentSessionService.java
siebeprojects: The network code has now been added to the exception message.

payment/src/main/java/net/optile/payment/core/Localization.java
siebeprojects: Key creation and definition of static key values have been moved to LocalizationKey class.

payment/src/main/java/net/optile/payment/core/Localization.java
siebeprojects: Yes, it is needed. At this stage, we only need to tell the Widget if it should show a little hint '?' icon. The translation of the hint is only needed when a PopupDialog is created by the presenter. The widget itself does not create the PopupDialog and thus does not need to store or have access to the actual translation of the popup message, image and button labels.

payment/src/main/java/net/optile/payment/ui/service/PaymentSessionService.java
siebeprojects: This is implemented in the PaymentSessionService where the ApplicableNetwork and Payment page properties files are loaded. If the new listUrl is different than the one used for loading the previous payment methods then the Localization will first be cleared.

PR #109, opened 2020-03-18
Pcx 428 change button label to ok

siebeprojects:


payment/src/main/res/values/strings.xml
```diff
@@ -19,5 +19,5 @@
     <string name="pmlocal_button_cancel">Cancel</string>
     <string name="pmlocal_button_retry">Retry</string>
     <string name="pmlocal_button_update">Update</string>
-    <string name="pmlocal_button_ok">OK</string>
+    <string name="pmlocal_button_ok">Ok</string>
```
hjhill: Are you sure it should be "Ok" and not "OK"? I've never seen it this way...

siebeprojects: According to my designer yes, it is "Ok". 
And so does this designer website use:
https://uxmovement.com/buttons/why-ok-buttons-in-dialog-boxes-work-best-on-the-right/

darklynx: @siebeprojects the very last comment of on the page refer is contradicting to yours statment :)

darklynx: Mac OS designers are also disagree: 
Screenshot 2020-03-18 at 20 33 03

darklynx: Also: https://www.google.com/search?q=iPhone+dialogs+with+OK&tbm=isch

PR #113, opened 2020-03-27
PCX-484: Refactor localisation in own service

siebeprojects:


payment/src/main/java/net/optile/payment/localization/Localization.java
```diff
@@ -44,179 +31,130 @@ public static Localization getInstance() {
...
-     * Helper method to obtain the translation for the error from the specified file.
+     * Helper method to obtain the translation for the given key from the localization file.
      *
-     * @param fileName name of the localization file
-     * @param error the error which should be translated
+     * @param networkCode name of the localization file
+     * @param key of the translation
      * @return the translation or null if not found
      */
-    public static String translateError(String fileName, String error) {
```
hjhill: @siebeprojects can you move the method `translateError` back here? This would reduce the number of changes, making reviewing simpler...

payment/src/main/java/net/optile/payment/ui/page/ChargePaymentPresenter.java
```diff
@@ -137,28 +143,59 @@ private void handleLoadSessionOk(PaymentSession session) {
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void onLocalizationError(Throwable cause) {
+        PaymentResult result = PaymentResult.fromThrowable(cause);
+        if (result.hasError()) {
+            handleLoadingError(result);
+        } else {
+            closeWithCanceledCode(result);
```
hjhill: A `PaymentResult.fromThrowable` that has no error? This is a strange concept...

payment/src/main/java/net/optile/payment/localization/LocalLocalizationHolder.java
```diff
@@ -0,0 +1,65 @@
...
+    private final Map<String, String> map;
+
+    /**
+     * Construct a new local localization holder
+     *
+     * @param context containing the local localizations
+     */
+    public LocalLocalizationHolder(Context context) {
+        this.map = new HashMap<>();
+        map.put(BUTTON_OK, context.getString(R.string.pmlocal_button_ok));
```
hjhill: Better use either `this.map` or `map`, not both.

payment/src/main/java/net/optile/payment/localization/Localization.java
siebeprojects: I try that but obviously not always successful. 
There are commits like PCX-484: Code cleanup and beautifying

payment/src/main/java/net/optile/payment/ui/page/ChargePaymentPresenter.java
siebeprojects: At the moment when we receive a 404 from the network, we do throw a PaymentException in the network class which ends up to be a Throwable in the service presenter. The error result of the 404 response may contain a JSON Object with Interaction Code. 

When we have a real error like IOException then result.hasError() would be true. But in any other case when we have a valid Interaction code received from the backend it is not treated as an internal SDK error and therefor result.hasError() will return false.

The next ticket PCX-500 will change this behaviour as we will be simplifying this error handling.

payment/src/main/java/net/optile/payment/localization/PropLocalizationHolder.java
```diff
@@ -0,0 +1,39 @@
...
+        this.properties = properties;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String translate(String key, String defValue) {
+        String value = properties.getProperty(key);
+        return TextUtils.isEmpty(value) ? defValue : value;
```
hjhill: This logic is repeated everywhere - wouldn't it be better to have a `translate(String key)` method (returning `null` if the key is not found) and introduce the default value elsewhere, for example
- in a level above the `LocalizationHolder`s 
- in a default `translate(String key, String defValue)` in the `LocalizationHolder` interface (slightly uglier than the other alternative, but still better than lots of repetition)?

payment/src/main/java/net/optile/payment/localization/Localization.java
```diff
@@ -10,30 +10,17 @@
...
 public final class Localization {
-    private final Map<String, Properties> files;
-    private LocalTranslations localTranslations;
-    private Properties sharedFile;
-    private String localizationId;
-
-    private Localization() {
-        this.files = new HashMap<>();
-    }
+    private Map<String, LocalizationHolder> networks;
```
hjhill: I have the feeling that re-using a single instance of this is not really worthwhile - could we make it immutable and use a new instance each time?

payment/src/main/java/net/optile/payment/ui/service/LocalizationService.java
```diff
@@ -0,0 +1,181 @@
...
+import net.optile.payment.localization.MultiLocalizationHolder;
+import net.optile.payment.network.LocalizationConnection;
+import net.optile.payment.ui.model.PaymentNetwork;
+import net.optile.payment.ui.model.PaymentSession;
+
+/**
+ * The LocalizationService providing asynchronize loading of the localizations needed for presenting the list of payment networks and showing errors.
+ * This service makes callbacks in the listener to notify of request completions.
+ */
+public final class LocalizationService {
```
hjhill: The way it is now, this is more of a `LocalizationContentLoaderService`.

payment/src/main/java/net/optile/payment/ui/page/ChargePaymentPresenter.java
hjhill: OK; otherwise I would have asked if this could be `hasInternalError` or `hasLoadingError` or something like this...

payment/src/main/java/net/optile/payment/localization/Localization.java
siebeprojects: I am leaving the single instance as it is for now. It has its advantages because setting holders in the Localization Object per payment session happens in two stages, 1. it is initialised with the local translations, 2 when we have loaded additional translations from the backend it is updated with the remote ones.

payment/src/main/java/net/optile/payment/localization/LocalLocalizationHolder.java
siebeprojects: this. is removed

payment/src/main/java/net/optile/payment/localization/PropLocalizationHolder.java
siebeprojects: The default value is now completely removed from Localization. 
Indeed it was not needed and not used by any classes in the SDK.

payment/src/main/java/net/optile/payment/localization/Localization.java
siebeprojects: The Localization class has now been made immutable and a new instance will be created each time. It is still required to have a current instance of the Localization class so this can now be set and retrieved through the get and set instance methods in the Localization class.

payment/src/main/java/net/optile/payment/localization/MultiLocalizationHolder.java
```diff
@@ -30,15 +30,13 @@ public MultiLocalizationHolder(LocalizationHolder... holders) {
...
      */
     @Override
-    public String translate(String key, String defValue) {
-        String value;
+    public String translate(String key) {
         for (LocalizationHolder holder : holders) {
-            value = holder.translate(key, null);
-
+            String value = holder.translate(key);
             if (!TextUtils.isEmpty(value)) {
```
hjhill, line 36: You should probably only check for `null` here.

payment/src/main/java/net/optile/payment/localization/Localization.java
```diff
@@ -154,9 +154,14 @@ public String getNetworkTranslation(String networkCode, String key, String defVa
      * @return the translation or defValue if the translation was not found
      */
     public String getSharedTranslation(String key, String defValue) {
-        return shared != null ? shared.translate(key, defValue) : defValue;
+        return getTranslationFromHolder(shared, key, defValue);
     }
 
+    private String getTranslationFromHolder(LocalizationHolder holder, String key, String defValue) {
+        String value = holder != null ? holder.translate(key) : null;
+        return TextUtils.isEmpty(value) ? defValue : value;
```
hjhill: You should probably only check for `null` here.

payment/src/main/java/net/optile/payment/localization/MultiLocalizationHolder.java
siebeprojects: It is a requirement to fall back to the next LocalizationHolder when the value is either null or empty.

PR #124, opened 2020-05-15
PCX-780: Refactor Expiry Date field

siebeprojects:


payment/src/main/java/net/optile/payment/ui/widget/DateWidget.java
```diff
@@ -72,113 +52,40 @@ public boolean validate() {
...
-                selYearIndex = i;
-            }
-        }
-        DateDialogFragment dialog = new DateDialogFragment();
-        dialog.setTitle(label);
-        dialog.setValues(selMonthIndex, monthLabels, selYearIndex, yearLabels);
-        dialog.setNeutralButton(dialogButtonLabel);
-        dialog.setListener(this);
-        return dialog;
+        int century = (Calendar.getInstance().get(Calendar.YEAR) / 100);
```
hjhill, line 78: Niggle: This mode of calculation will break with the first use of a credit card valid until 2100 or longer. 😉

payment/src/main/java/net/optile/payment/ui/widget/TextInputWidget.java
```diff
@@ -11,21 +11,19 @@
...
+import net.optile.payment.ui.widget.input.EditTextInputMode;
+import net.optile.payment.ui.widget.input.EditTextInputModeFactory;
 import net.optile.payment.validation.ValidationResult;
 
 /**
  * Widget for handling text input
  */
-public final class TextInputWidget extends InputLayoutWidget {
-    private TextInputMode mode;
+public class TextInputWidget extends InputLayoutWidget {
```
hjhill, line 26: Why was `final` removed?

payment/src/main/java/net/optile/payment/ui/widget/input/EditTextWatcher.java
```diff
@@ -0,0 +1,88 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public void afterTextChanged(final Editable editable) {
+        if (internalLock) {
+            return;
+        }
+        internalLock = true;
+        validate(editable);
+        internalLock = false;
```
hjhill, line 67:
```suggestion
        try {
            validate(editable);
        } finally {
            internalLock = false;
        }
```

payment/src/main/java/net/optile/payment/ui/widget/TextInputWidget.java
siebeprojects: It can be put back, during implementation I made DateWidget extend from TextInputWidget but this is not the case anymore. I will put it back.

payment/src/main/java/net/optile/payment/ui/widget/input/EditTextWatcher.java
siebeprojects: Yes, that is a better way, will change it.

payment/src/main/java/net/optile/payment/ui/widget/input/EditTextWatcher.java
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Base class for EditText watchers
+ */
+public class EditTextWatcher implements TextWatcher {
+
+    final EditText editText;
+    String beforeText;
+    int beforeStart;
+    int beforeCount;
+    int beforeAfter;
```
hjhill, line 24: `beforeAfter` is a super-strange name. This is the length of the replacement text, right? (It might be best to change some of the other names beginning with `before...` to have a consistent naming that conveys the meaning of all of those fields).

payment/src/main/java/net/optile/payment/ui/widget/input/GroupingTextWatcher.java
```diff
@@ -4,42 +4,33 @@
...
     private static final String STRING_DIVIDER = " ";
 
-    private final EditText editText;
     private final int groupSize;
     private final int dividerModulo;
     private final String inputRegex;
     private final Pattern inputPattern;
 
-    private String previousText = "";
-    private int deleteLength;
```
hjhill, line 25: I like those names much better than `before...` (especially `beforeAfter`).

payment/src/main/java/net/optile/payment/validation/Validator.java
```diff
@@ -36,6 +36,8 @@
     public final static int MAXLENGTH_IBAN = 34;
     public final static int MAXLENGTH_BIC = 11;
 
+    public final static int MAX_EXPIRY_YEAR = 15;
```
hjhill, line 39: Caution! There seem to be credit cards with much longer expiration dates - see the last sentence of https://stackoverflow.com/a/2500654/32316...

payment/src/main/java/net/optile/payment/ui/widget/input/EditTextWatcher.java
```diff
@@ -0,0 +1,88 @@
...
+    boolean internalLock;
+
+    public EditTextWatcher(EditText editText) {
```
hjhill, line 27: Do you really need the reference to the `EditText`? What is it used for?

payment/src/main/java/net/optile/payment/ui/widget/input/ExpiryDateInputMode.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void apply(TextInputEditText editText) {
+        InputFilter[] filters = new InputFilter[1];
+        filters[0] = new InputFilter.LengthFilter(maxLength);
+        editText.setFilters(filters);
```
hjhill, line 40: Same number of lines but maybe a bit easier to read:
```suggestion
        editText.setFilters(new InputFilter[] {
            new InputFilter.LengthFilter(maxLength);
        });
```

payment/src/main/java/net/optile/payment/ui/widget/DateWidget.java
siebeprojects: This is now fixed with windowing of -30 and +70 years.

payment/src/main/java/net/optile/payment/ui/widget/input/EditTextWatcher.java
siebeprojects: Added the try finally catch to reset the lock in case of an exception

payment/src/main/java/net/optile/payment/ui/widget/input/ExpiryDateInputMode.java
siebeprojects: Changed this in the ExpiryDateInputMode and also in the other modes.

payment/src/main/java/net/optile/payment/ui/widget/input/EditTextWatcher.java
siebeprojects: yes, both children ExpiryDateTextWatcher and GroupingTextWatcher need the EditText reference to manipulate the cursor position in the EditText. Since you cannot do that through the Editable. I didnt want the TextWatcher to have a reference to the InputMode.

payment/src/main/java/net/optile/payment/ui/widget/input/GroupingTextWatcher.java
siebeprojects: Changed the names back to the original ones

payment/src/main/java/net/optile/payment/validation/Validator.java
siebeprojects: Changed the expiration date to 50 years

payment/src/main/java/net/optile/payment/ui/widget/input/EditTextWatcher.java
siebeprojects: Changed the names back to the previous names

payment/src/main/java/net/optile/payment/ui/widget/input/EditTextWatcher.java
hjhill: Good point - OK, we'll have to live with that unless we want to do a big refactoring on the duplicate class hierarchy.

PR #126, opened 2020-05-28
PCX-840: Obtain NetworkService through code and method

siebeprojects:


payment/src/main/java/net/optile/network/basic/BasicNetworkServiceFactory.java
```diff
@@ -23,42 +24,23 @@
      * {@inheritDoc}
      */
     @Override
-    public boolean isNetworkSupported(ApplicableNetwork network) {
-        String code = network.getCode();
-
-        // This service only supports redirect Paypal
-        if (PaymentNetworkCodes.PAYPAL.equals(code)) {
-            return network.getRedirect();
+    public boolean isSupported(String code, String method) {
```
hjhill, line 27: As you're already refactoring: what do you think about naming this method `supports`? It would make for a more natural reading when called and it does not have to adhere to the Bean getter naming convention...

payment/src/main/java/net/optile/payment/model/AccountRegistration.java
```diff
@@ -71,6 +76,43 @@ public void setCode(final String code) {
...
+     * @param method the method to set.
+     */
+    public void setMethod(@PaymentMethod.Definition final String method) {
+        this.method = method;
+    }
+
+    /**
+     * Gets code of button's label what should be used if this network is selected.
+     *
+     * @return Code of button'S label.
```
hjhill, line 101: small typo: "button's" (also below)

payment/src/main/java/net/optile/payment/ui/service/NetworkServiceLookup.java
```diff
@@ -10,46 +10,51 @@
...
+     * Lookup a NetworkService for the network code and payment method
      *
-     * @param code the applicable network code
+     * @param code to be used to lookup a NetworkService
+     * @param method to be used to lookup a NetworkService
      * @return the NetworkService that can handle the network or null if none found
      */
-    public static NetworkService getService(String code) {
-        NetworkServiceFactory factory = getFactory(code);
+    public static NetworkService getService(String code, String method) {
```
hjhill, line 44: This should actually be `createService` as it creates a new object.

siebeprojects: Changed to createService

payment/src/main/java/net/optile/payment/model/AccountRegistration.java
siebeprojects: removed the 's behind button

payment/src/main/java/net/optile/network/basic/BasicNetworkServiceFactory.java
siebeprojects: changed to supports()

PR #133, opened 2020-06-26
PCX-768: Show both title and message in Message Dialogs

siebeprojects:
This required refactoring of the BasePaymentActivity and PaymentViews

### DIFF
--- a/payment/src/main/java/net/optile/payment/ui/page/BasePaymentActivity.java
+++ b/payment/src/main/java/net/optile/payment/ui/page/BasePaymentActivity.java
@@ -30,7 +30,7 @@ import net.optile.payment.ui.page.idlingresource.SimpleIdlingResource;
 /**
  * The base activity for payment activities.
  */
-abstract class BasePaymentActivity extends AppCompatActivity implements BasePaymentView {
+abstract class BasePaymentActivity extends AppCompatActivity implements PaymentView {
 
     boolean active;
     ProgressView progressView;
--- a/payment/src/main/java/net/optile/payment/ui/page/BasePaymentActivity.java
+++ b/payment/src/main/java/net/optile/payment/ui/page/BasePaymentActivity.java
@@ -8,33 +8,37 @@
 
 package net.optile.payment.ui.page;
 
-import static net.optile.payment.localization.LocalizationKey.BUTTON_CANCEL;
-import static net.optile.payment.localization.LocalizationKey.BUTTON_OK;
-import static net.optile.payment.localization.LocalizationKey.BUTTON_RETRY;
-import static net.optile.payment.localization.LocalizationKey.ERROR_CONNECTION_TEXT;
-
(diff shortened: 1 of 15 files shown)


payment/src/main/res/values/strings.xml
```diff
@@ -3,9 +3,11 @@
     <string name="pmpage_error_notsupported">The payment methods in the current payment session are not supported by this app.</string>
     <string name="pmpage_contentdescription">PaymentImage</string>
 
+    <string name="error_connection_title">Oops!</string>
     <string name="error_connection_text">There was a problem with your internet connection</string>
+    <string name="error_default_title">Oops!</string>
     <string name="error_default_text">An error occurred while handling your payment. Please try again.</string>
-
+
```
hjhill, line 10: Oops.

payment/src/main/java/net/optile/payment/ui/page/PaymentListPresenter.java
```diff
@@ -528,25 +518,22 @@ public void onButtonClicked(ThemedDialogFragment dialog, int which) {
             public void onDismissed(ThemedDialogFragment dialog) {
                 view.close();
             }
-        });
+        };
+        if (Localization.hasInteraction(interaction)) {
+            view.showInteractionDialog(interaction, listener);
+        } else {
+            view.showDefaultErrorDialog(listener);
+        }
```
hjhill, line 526: This is a pattern that occurs more than once and feels a bit bad to me because of split responsibility - would it be better to push the responsibility for handling "Localization not found" to the component that retrieves the localization (i.e., the `view`)?

siebeprojects: in an MVP the presenter decides what the View should show, the View aught to be passive. Now, I do think we could move shared functionality between the two presenters (list and charge) to a shared class. But for that it would be nice we would have a whiteboard sometimes and sit together.

payment/src/main/res/values/strings.xml
siebeprojects: Will fix it

payment/src/main/java/net/optile/payment/ui/page/BasePaymentActivity.java
```diff
@@ -8,34 +8,38 @@
...
-abstract class BasePaymentActivity extends AppCompatActivity {
+abstract class BasePaymentActivity extends AppCompatActivity implements PaymentView {
 
     boolean active;
     ProgressView progressView;
 
+    /** For testing only */
+    SimpleIdlingResource closeIdlingResource;
+    SimpleIdlingResource dialogIdlingResource;
+    boolean closed;
```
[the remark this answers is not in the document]
siebeprojects: We did not need the chargecompleted idling resource anymore and can re-use the same CloseIdlingResource.

PR #134, opened 2020-06-30
PCX-767: Update register/recurrence checkboxes

siebeprojects:


payment/src/main/java/net/optile/payment/ui/widget/CheckBoxWidget.java
```diff
@@ -64,9 +65,10 @@ boolean isChecked() {
         return value.isChecked();
     }
 
-    void initCheckBox(boolean clickable, boolean checked) {
+    void initCheckBox(boolean visible, boolean clickable, boolean checked) {
```
hjhill, line 68: With the current changes, you don't really use `clickable` anymore (only `false` when it is hidden). The `boolean visible` also bothers me a bit because:
- you need a ternary operator to convert it to the value used in `setVisible` and
- if it is set to `false`, it makes the other parameters meaningless.

Maybe having two separate methods `initCheckBox(boolean checked)` and `hideCheckBox()` would be better?

siebeprojects: Agree, I separated the logic of the init method to setCheckboxVisible and setChecked. 
Indeed the setClickable is not needed anymore as it can always be clickable since the checkbox will be hidden in FORCED mode.

PR #139, opened 2020-07-09
PCX-916: Remove all custom theming and deprecated features

siebeprojects:


payment/src/main/java/net/optile/payment/ui/page/ProgressView.java
```diff
@@ -38,7 +34,7 @@
         progressBar = view.findViewById(R.id.progressbar);
         textHeader = view.findViewById(R.id.text_header);
         textInfo = view.findViewById(R.id.text_info);
-        styleProgressBar();
+        //styleProgressBar();
```
hjhill, line 37: Can we get rid of this code instead of commenting it out?

siebeprojects: It seems material.io does not support progressbars for Android yet. I enabled the custom styling again.

PR #144, opened 2020-07-16
PCX-785: Update form fields according to new design

siebeprojects:
Left out from this PR:
Small logos in NetworkCards for smartswitch
Delete button in TextInput fields to delete input fields


payment/src/main/java/net/optile/payment/ui/list/NetworkCardViewHolder.java
```diff
@@ -56,57 +58,54 @@ static ViewHolder createInstance(ListAdapter adapter, NetworkCard networkCard, V
...
         bindRecurrenceWidget(network);
     }
 
-    private void addRegisterWidgets(PaymentTheme theme) {
-        FormWidget widget = WidgetInflater.inflateRegisterWidget(AUTO_REGISTRATION, formLayout, theme);
-        addWidget(widget);
-        widget = WidgetInflater.inflateRegisterWidget(ALLOW_RECURRENCE, formLayout, theme);
-        addWidget(widget);
-    }
+    private void bindGroupedPaymentNetworks(NetworkCard networkCard) {
```
hjhill, line 85: A lot of the contents of this method is duplicated from `bindSinglePaymentNetwork` - can we structure it so that we have as few duplications as possible?

siebeprojects: Restructured this with limited duplications.

payment/src/main/java/net/optile/payment/ui/widget/InputLayoutWidget.java
```diff
@@ -125,7 +173,8 @@ void setTextInputMode(EditTextInputMode mode) {
 
     String getValue() {
         CharSequence cs = textInput.getText();
-        return cs != null ? cs.toString().trim() : "";
+        String val = cs != null ? cs.toString().trim() : "";
+        return mode != null ? mode.normalize(val) : val;
```
hjhill, line 177: Brackets would make this a bit more readable:
```suggestion
        String val = (cs != null) ? cs.toString().trim() : "";
        return (mode != null) ? mode.normalize(val) : val;
```

payment/src/main/java/net/optile/payment/ui/widget/WidgetPresenter.java
```diff
@@ -38,6 +38,15 @@
...
     void showKeyboard();
 
+    /**
+     * Get the max length for the given input type
+     *
+     * @param code of the network
+     * @param type of the input
+     * @return the max length or -1 if not available
+     */
+    int getMaxLength(String code, String type);
```
hjhill, line 48: Recommendation: Better use `networkCode` (here and in the implementations) as `code` is too generic.

payment/src/main/java/net/optile/payment/ui/list/NetworkCardViewHolder.java
```diff
@@ -56,57 +49,44 @@ static ViewHolder createInstance(ListAdapter adapter, NetworkCard networkCard, V
...
+            logoView.setImageResource(R.drawable.ic_card);
+            bindPaymentNetwork("group", network);
+        }
+    }
+
+    private void bindPaymentNetwork(String testId, PaymentNetwork network) {
+        PaymentUtils.setTestId(itemView, "card", testId);
         bindRegistrationWidget(network);
         bindRecurrenceWidget(network);
     }
```
hjhill, line 74: This functional separation still feels a bit "uneven" - what do you think about this one:
```suggestion
        if (networkCard.getPaymentNetworkSize() == 1) {
            bindLogoView(networkCard.getCode(), networkCard.getLink("logo"));
            setTestId("network");
        } else {
            logoView.setImageResource(R.drawable.ic_card);
            setTestId("group");
        }

        bindRegistrationWidget(network);
        bindRecurrenceWidget(network);
    }

    private void setTestId(String testId) {
        PaymentUtils.setTestId(itemView, "card", testId);
    }
```

payment/src/main/java/net/optile/payment/ui/list/PaymentCardViewHolder.java
```diff
@@ -95,37 +87,35 @@ public FormWidget getFormWidget(String name) {
...
-                case PaymentInputType.EXPIRY_YEAR:
-                    if (dateWidget == null) {
-                        dateWidget = WidgetInflater.inflateDateWidget(PaymentInputType.EXPIRY_DATE, formLayout, theme);
-                        addWidget(dateWidget);
+                case VERIFICATION_CODE:
+                    addWidget(WidgetInflater.inflateVerificationCodeWidget(VERIFICATION_CODE, formLayout));
+                    break;
+                case EXPIRY_MONTH:
+                case EXPIRY_YEAR:
+                    if (!containsExpiryDate) {
```
hjhill, line 111: Minor: A more expressive name like `elementsContainsExpiryDate` would make understanding slightly easier.

payment/src/main/java/net/optile/payment/ui/list/NetworkCardViewHolder.java
siebeprojects: yes, that is a lot better, will change it.

payment/src/main/java/net/optile/payment/ui/widget/WidgetPresenter.java
siebeprojects: I agree, but let's leave it in this PR. If we change this here we need to change it everywhere in the code and not only at one or two places. This would be better in a separate PR.

payment/src/main/java/net/optile/payment/ui/widget/InputLayoutWidget.java
siebeprojects: Fixed

payment/src/main/java/net/optile/payment/ui/list/PaymentCardViewHolder.java
siebeprojects: fixed

PR #147, opened 2020-07-27
PCX-786: Replace custom dialogs with material alert dialogs

siebeprojects:


payment/src/main/java/net/optile/payment/ui/dialog/PaymentDialogFragment.java
```diff
@@ -0,0 +1,173 @@
...
+     * {@inheritDoc}
+     */
+    @NonNull
+    @Override
+    public Dialog onCreateDialog(Bundle savedInstanceState) {
+        Activity activity = getActivity();
+        if (activity == null) {
+            return super.onCreateDialog(savedInstanceState);
+        }
+        LayoutInflater inflater = activity.getLayoutInflater();
```
hjhill, line 110: You don't seem to use `inflater`.

siebeprojects: Removed this variable.

PR #149, opened 2020-07-29
PCX-957: Add network logos to grouped card

siebeprojects:


payment/src/main/java/net/optile/payment/ui/model/SmartSwitch.java
```diff
@@ -66,6 +66,15 @@ public boolean hasSelected() {
         return smartSelected.size() > 0;
     }
 
+    /**
+     * Get the number of selected networks
+     *
+     * @return the number of selected networks
+     */
+    public int getSelectedSize() {
```
hjhill, line 74: `Size` reads a bit strange in that context - better name it `getSelectedCount`

payment/src/main/java/net/optile/payment/ui/list/NetworkCardViewHolder.java
```diff
@@ -59,16 +62,37 @@ void onBind(PaymentCard paymentCard) {
...
+        if (smartSwitch.getSelectedSize() == 1) {
+            PaymentNetwork network = smartSwitch.getFirstSelected();
+            networkLogosView.setSelected(network.getCode());
+            return;
+        }
+        networkLogosView.setSelected(null);
+    }
+
+    private void addNetworkLogos(View parent, NetworkCard networkCard) {
+        if (networkCard.getPaymentNetworkSize() <= 1) {
```
hjhill, line 90: This is another case of `Size` where `Count` would probably be better - or

payment/src/main/java/net/optile/payment/ui/model/SmartSwitch.java
siebeprojects: Changed to count

payment/src/main/java/net/optile/payment/ui/list/NetworkCardViewHolder.java
siebeprojects: Changed as well to count

PR #152, opened 2020-07-31
PCX-935: Update example apps with material UI

siebeprojects:


example-basic/src/main/java/net/optile/example/basic/SdkThemeBuilder.java
```diff
@@ -31,6 +31,8 @@ public static PaymentTheme createDefaultTheme() {
      */
     public static PaymentTheme createCustomTheme() {
         return PaymentTheme.createBuilder().
+            setPaymentListTheme(R.style.CustomTheme_Toolbar).
+            setChargePaymentTheme(R.style.CustomTheme).
             build();
```
hjhill, line 36: Small improvement of readability: starting a new line with `.` shows immediately that this is a continuation:
```suggestion
        return PaymentTheme.createBuilder()
            .setPaymentListTheme(R.style.CustomTheme_Toolbar)
            .setChargePaymentTheme(R.style.CustomTheme)
            .build();
```

PR #157, opened 2020-08-12
PCX-999: Fix input focus for preselected cards

siebeprojects:


payment/src/main/java/net/optile/payment/ui/list/PaymentList.java
```diff
@@ -182,6 +175,16 @@ private int nextViewType() {
...
     }
 
+    private void setEmptyMessage(PaymentSession session) {
+        String msg = "";
+        if (session.getApplicableNetworkSize() == 0) {
+            msg = activity.getString(R.string.pmpage_error_empty);
+        } else if (session.getNetworkCardSize() == 0) {
+            msg = activity.getString(R.string.pmpage_error_notsupported);
+        }
+        emptyMessage.setText(msg);
```
hjhill, line 185: Minor: I'd have a slight preference for pushing the `emptyMessage.setText` to the caller - then this could look like:

```
    private String getEmptyMessageText(PaymentSession session) {
        if (session.getApplicableNetworkSize() == 0) {
            return activity.getString(R.string.pmpage_error_empty);
        } else if (session.getNetworkCardSize() == 0) {
            return activity.getString(R.string.pmpage_error_notsupported);
        } else {
            return "";
        }
```

and the caller would do `emptyMessage.setText(getEmptyMessageText(session))`.

PR #158, opened 2020-08-12
PCX-1000: Fix crash in SummaryActivity

siebeprojects:
SummaryActivity may crash when there was no payment network for the stored preset account.

### DIFF
--- a/example-demo/src/main/java/net/optile/example/demo/summary/SummaryPresenter.java
+++ b/example-demo/src/main/java/net/optile/example/demo/summary/SummaryPresenter.java
@@ -161,12 +161,7 @@ final class SummaryPresenter {
         this.subscription = null;
         this.result = result;
         PresetAccount account = result.getPresetAccount();
-
-        if (account == null) {
-            view.closeScreen();
-            return;
-        }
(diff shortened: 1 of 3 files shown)


example-demo/src/main/java/net/optile/example/demo/summary/SummaryPresenter.java
```diff
@@ -154,7 +161,17 @@ private void callbackLoadPaymentSessionSuccess(ListResult result) {
         this.subscription = null;
         this.result = result;
         PresetAccount account = result.getPresetAccount();
-        view.showPaymentDetails(account, getPaymentMethod(account.getCode(), result));
+
+        if (account == null) {
+            view.closeScreen();
```
hjhill, line 166: I don't like the duplication - I think moving the null-check for `account` (and `account.getCode()`) into `getPaymentMethod` would improve the code a bit.

siebeprojects: Good point, I moved the null check to getPaymentMethod. The Payment API is being modified whereby the PresetAccount model class will also contain the paymentMethod. This will make the entire getPaymentMethod() method finally obsolete.

PR #160, opened 2020-08-13
PCX-937: Change CANCEL flow to ERROR flow

siebeprojects:


example-demo/src/main/java/net/optile/example/demo/summary/SummaryPresenter.java
```diff
@@ -154,7 +161,17 @@ private void callbackLoadPaymentSessionSuccess(ListResult result) {
         this.subscription = null;
         this.result = result;
         PresetAccount account = result.getPresetAccount();
-        view.showPaymentDetails(account, getPaymentMethod(account.getCode(), result));
+
+        if (account == null) {
+            view.closeScreen();
+            return;
+        }
+        String paymentMethod = getPaymentMethod(account.getCode(), result);
```
hjhill, line 169: Minor: Moving the `null`-check for `account` into `getPaymentMethod` could make the code a bit cleaner here - you could use `if (paymentMethod==null) view.closeScreen(); else view.showPaymentDetails(...);` then.

payment/src/main/java/net/optile/payment/ui/page/BasePaymentActivity.java
```diff
@@ -123,6 +111,10 @@ public void showInteractionDialog(Interaction interaction, PaymentDialogListener
         if (!active) {
             return;
         }
+        if (!Localization.hasInteraction(interaction)) {
+            showDefaultErrorDialog(listener);
+            return;
+        }
         progressView.setVisible(false);
```
hjhill, line 118: Code duplication: this is done in `showDefaultErrorDialog` as well, maybe move it above the `if`.

payment/src/main/java/net/optile/payment/ui/page/BasePaymentActivity.java
```diff
@@ -189,6 +181,12 @@ public void showDialogFragment(DialogFragment dialog, String tag) {
         }
     }
 
+    void showDefaultErrorDialog(PaymentDialogListener listener) {
```
hjhill, line 184: Why does this have to be a separate method? Using only `PaymentDialogFragment` instead of `DialogFragment` and moving the "tag" as an attribute into that class might help to make it smaller so that it can be inlined again.

example-demo/src/main/java/net/optile/example/demo/summary/SummaryPresenter.java
siebeprojects: The null check is moved to getPaymentMethod.

payment/src/main/java/net/optile/payment/ui/page/BasePaymentActivity.java
siebeprojects: Refactored the creation and showing of PaymentDialogs.

payment/src/main/java/net/optile/payment/ui/page/BasePaymentActivity.java
siebeprojects: This separation has been removed and a "tag" attribute has been added to the PaymentDialogFragment class.

hjhill (approved):

PR #167, opened 2020-09-16
PCX-1025: Refactor PaymentResult and deprecate PaymentError

siebeprojects:
**Description**
As a merchant developer, the PaymentResult should not contain duplicated Interaction + resultInfo information which may confuse me.

Currently the PaymentResult contains an Interaction Object + resultInfo that are copied from the also included OperationResult.

**Acceptance Criteria**

- [x] Use the ErrorInfo object to store Interaction + resultInfo if there is no OperationResult in the PaymentResult

- [x] Return the ErrorInfo from e.g. the 404 list response in the PaymentResult

** Resolution **
The PaymentError class has been removed and only the PaymentResult class is used to store the OperationResult, ErrorInfo and optional Throwable. 
Also the onPrepare stage for processing a payment has been removed as this was using the PaymentResult which should not be used for the prepare stage.


example-basic/src/main/java/net/optile/example/basic/BasicActivity.java
```diff
@@ -99,18 +100,19 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
         if (requestCode != PAYMENT_REQUEST_CODE) {
             return;
         }
-        PaymentResult paymentResult = PaymentResult.fromResultIntent(data);
+        PaymentResult paymentResult = PaymentResultHelper.fromResultIntent(data);
         if (paymentResult != null) {
-            this.sdkResult = new SdkResult(resultCode, paymentResult);
+            sdkResult = new PaymentActivityResult(requestCode, resultCode, paymentResult);
         }
```
hjhill, line 106: Two things:
1. `Do you expect clients to have `PaymentActivityResult` instances with different `requestCode` values? From all example code I've seen, I don't think `PaymentActivityResult.requestCode` will ever be used.
2. What do you think about adding a static factory method `PaymentActivityResult.fromCodeAndIntent(resultCode, paymentResult)` (or `...(requestCode, resultCode, paymentResult)` if keeping `requestCode` makes sense) that would do the low-level work here?

example-basic/src/main/java/net/optile/example/basic/BasicActivity.java
```diff
@@ -99,18 +100,19 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
-    private void showSdkResult(SdkResult sdkResult) {
-        PaymentResult paymentResult = sdkResult.paymentResult;
+    private void showSdkResult(PaymentActivityResult sdkResult) {
+        PaymentResult paymentResult = sdkResult.getPaymentResult();
+        int resultCode = sdkResult.getResultCode();
 
         resultHeaderView.setVisibility(View.VISIBLE);
         resultLayout.setVisibility(View.VISIBLE);
-        setText(resultCodeView, sdkResult.getResultCodeString());
+        setText(resultCodeView, PaymentActivityResult.resultCodeToString(resultCode));
```
hjhill, line 115: Separation of SDK and client: Is `resultCodeToString` something a real client would use? If no, I would move this logic into the example app.

example-demo/src/main/java/net/optile/example/demo/checkout/CheckoutActivity.java
```diff
@@ -129,9 +130,9 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
         if (requestCode != PAYMENT_REQUEST_CODE) {
             return;
         }
-        PaymentResult result = PaymentResult.fromResultIntent(data);
-        if (result != null) {
-            this.sdkResult = new SdkResult(requestCode, resultCode, result);
+        PaymentResult paymentResult = PaymentResultHelper.fromResultIntent(data);
+        if (paymentResult != null) {
+            sdkResult = new PaymentActivityResult(requestCode, resultCode, paymentResult);
```
hjhill, line 135: Don't you have that already in the base class?

payment/src/main/java/net/optile/payment/ui/PaymentResult.java
```diff
@@ -39,138 +37,105 @@ public PaymentResult createFromParcel(Parcel in) {
             return new PaymentResult[size];
         }
     };
-    private String resultInfo;
-    private Interaction interaction;
+
+    /**
+     * Provides detailed information about the performed operation
```
hjhill, line 42:
```suggestion
     * Provides detailed information about the result of the performed operation
```

hjhill: Another thing: Wouldn't it make sense to use the one-line form of Javadoc comments for fields?

payment/src/main/java/net/optile/payment/ui/PaymentResult.java
```diff
@@ -208,18 +163,13 @@ public void writeToParcel(Parcel out, int flags) {
...
-            sb.append(interaction.getReason());
-        }
-        if (this.error != null) {
-            sb.append(error.toString());
-        }
+        sb.append(getResultInfo());
+        sb.append(", code: ");
+        sb.append(interaction.getCode());
+        sb.append(", reason: ");
+        sb.append(interaction.getReason());
```
hjhill, line 172: If you don't use any `if` statements, using concatenation directly does the same and is easier to read:

`return "PaymentResult[resultInfo: " + getResultInfo() + ...`

example-basic/src/main/java/net/optile/example/basic/BasicActivity.java
siebeprojects: The requestCode is used, for example in the Summary screen where we make two different requests. 1. to edit the preset account and 2. to charge a preset account. The requestCode is then used to identify to which action the activity result belongs to. 

The low-level work is only setting the class members (requestCode, responseCode and PaymentResult) in the container class. I do not see the benefit in this case to make a static function to replace a simple constructor that will do exactly the same.

example-basic/src/main/java/net/optile/example/basic/BasicActivity.java
siebeprojects: True, but this class is also used inside the Android-SDK and while developing the Android SDK this is handy to have to print out debug information of what the resultCode is.

example-demo/src/main/java/net/optile/example/demo/checkout/CheckoutActivity.java
siebeprojects: Moved the onActivityResult handling to the Base class.

payment/src/main/java/net/optile/payment/ui/PaymentResult.java
siebeprojects: Removed the comments from private class members in PaymentResult

payment/src/main/java/net/optile/payment/ui/PaymentResult.java
siebeprojects: Removed using the StringBuilder.

example-basic/src/main/java/net/optile/example/basic/BasicActivity.java
hjhill: I agree that it should be in the SDK in that case, but having `resultCodeToString` in a different class than the constants for `resultCode` feels wrong. 

I recommend to move them to the same class, maybe even create a small (sub)class only for the result code and the `toString` logic.

siebeprojects: Well, it is done the same as it is done in the Android class which represents the ActivityResult. If they have it like this then it is acceptable to me, neither did Android put it into the Activity class. 

https://developer.android.com/reference/androidx/activity/result/ActivityResult

hjhill (approved):

PR #168, opened 2020-09-28
PCX-841: Use single account registration

siebeprojects:
This PR removes that AccountRegistration Objects must have an ApplicableNetwork in the same ListResult.


payment/src/main/java/net/optile/payment/model/PresetAccount.java
```diff
@@ -95,4 +98,23 @@ public AccountMask getMaskedAccount() {
...
     }
+
+    /**
+     * Gets value of method.
+     *
+     * @return the method.
+     */
+    @PaymentMethod.Definition
+    public String getMethod() {
+        return method;
```
bzlat, line 109: Why not name the field `paymentMethod` => `getPaymentMethod` => disambiguating exactly what `method` is?

payment/src/main/java/net/optile/payment/model/PresetAccount.java
```diff
@@ -95,4 +98,23 @@ public AccountMask getMaskedAccount() {
     public void setMaskedAccount(final AccountMask maskedAccount) {
         this.maskedAccount = maskedAccount;
     }
+
+    /**
+     * Gets value of method.
```
bzlat, line 103: Is this `javadoc` generated automatically?

payment/src/main/java/net/optile/payment/model/PresetAccount.java
siebeprojects: These model classes are a copy from the client API model classes from Optile that can be downloaded from optile.io. 
For now, we still keep them the same.

payment/src/main/java/net/optile/payment/model/PresetAccount.java
siebeprojects: Same here, these classes are cloned from the model classes provided on optile.io and thus also copy the javadocs defined in the publicly available model classes.

example-demo/src/main/java/net/optile/example/demo/summary/SummaryActivity.java
```diff
@@ -162,7 +161,7 @@ public void showPaymentDetails(PresetAccount presetAccount, String method) {
         }
         presetSubtitle.setVisibility(View.GONE);
         if (mask != null) {
-            setAccountMask(mask, method);
+            setAccountMask(mask, presetAccount.getMethod());
```
hjhill, line 164: Good catch! 👍

example-demo/src/main/java/net/optile/example/demo/summary/SummaryPresenter.java
```diff
@@ -106,12 +106,12 @@ public void onError(Throwable error) {
...
         switch (result.getResultCode()) {
-            case RESULT_CODE_PROCEED:
-                loadPaymentDetails(view.getListUrl());
-                break;
             case RESULT_CODE_ERROR:
                 handlePaymentResultError(result.getPaymentResult());
                 break;
+            default:
+                loadPaymentDetails(view.getListUrl());
+                break;
```
hjhill, line 114: May I ask why this was changed? Code checker warning about "no default"?

siebeprojects: In preset flow the situation may arise that the preset account it removed by the OPG system after a failed charge request on an applicable network. So, in this case it may happen that the payment list is reopened by the merchant and the user should preset a new payment method because the previously one has been removed by the OPG System.

The user is allowed to close the payment list without presetting another network. We decided that when the user performs no action at all in the payment list, we do not return a PaymentResult. (Android returns Activity.RESULT_CANCELED as the resultCode in this case).

The Summary page is having control again and it must check if the current ListResult contains a PresetAccount. If the ListResult does not contain a PresetAccount then it will close itself and reopen the checkout page.

example-demo/src/main/java/net/optile/example/demo/summary/SummaryActivity.java
siebeprojects: Yes, finally the PresetAccount has the method field so we do not need the ApplicableNetworks anymore. 
Big improvement and we can finally remove the code that obtained the method from any of the ApplicableNetworks.

example-demo/src/main/java/net/optile/example/demo/summary/SummaryPresenter.java
hjhill: Your explanation makes sense, but the reasoning is IMO not obvious. What do you think about replacing the `switch` with an `if/else` and adding a comment about this? It could be similar to this:
```
if (result.getResultCode() == RESULT_CODE_ERROR) {
    handlePaymentResultError(result.getPaymentResult());
} else {
    // result could be RESULT_CODE_PROCEED or Activity.RESULT_CANCELED (user closed payment list without edits)
    loadPaymentDetails(view.getListUrl());
}
```

hjhill (approved):

PR #174, opened 2020-10-13
test: Merge

arenstar:


siebeprojects:
This PR should not be needed, the problem of packagecloud is with the packagecloud user that does have the proper rights to publish to packagecloud.

PR #180, opened 2020-10-30
PCX-961: Only use localization files for labels

siebeprojects:


payment/src/main/java/net/optile/payment/ui/model/AccountCard.java
```diff
@@ -79,7 +79,7 @@ public String getLabel() {
      */
     @Override
     public List<InputElement> getInputElements() {
-        List<InputElement> elements = account.getLocalizedInputElements();
+        List<InputElement> elements = account.getInputElements();
         return elements == null ? new ArrayList<>() : elements;
```
oxelad, line 83: I think it is better to return `Collections.empyList()` instead of `new ArrayList<>()` (here and in `PaymentNetwork` and may be other classes)

bzlat: Unless it's used like:
```
AccountCard accoundCard = new AccountCard();
accountCard.getInputElements().add(...)

or

List<InputElements> accountCardInputElements = accountCard.getInputElements();
accountCardInputElements.add(...)
```

siebeprojects: @bzlat View model classes e.g. AccountCard do not store data themselves but instead provide access to data directly from the model classes obtained from the OPG Payment API. In that sense the AccountCard view model is a wrapper for the AccountRegistration obtained from the backend. Benefit was that we did not need to copy lots of data from the server model classes into the View model classes, but obviously a disadvantage is that we would deal more with e.g. the List<InputElement> format as it is stored in the server model class AccountRegistration.

bzlat (approved):

PR #183, opened 2020-11-10
PCX-730: Support POST redirects

siebeprojects:


siebeprojects:
This ticket can be tested with browserstack app: Example-SDK-PCX-730-debug.apk

payment/src/main/java/net/optile/payment/ui/redirect/RedirectRequest.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.net.URL;
+import java.util.Map;
+
+import net.optile.payment.core.PaymentException;
+import net.optile.payment.model.OperationResult;
+import net.optile.payment.model.Redirect;
+
+/**
+ * Class holding the data for making a Request request
```
hjhill, line 19: oops

```suggestion
 * Class holding the data for making a Redirect request
```

payment/src/main/java/net/optile/payment/ui/redirect/RedirectService.java
```diff
@@ -31,28 +33,35 @@
...
-    public static void open(Context context, Redirect redirect) throws PaymentException {
-        if (!isSupported(context, redirect)) {
-            throw new PaymentException("Redirect payment is not supported by the Android-SDK");
+    public static void redirect(Context context, RedirectRequest request) throws PaymentException {
+        if (!supports(context, request)) {
+            throw new PaymentException("The redirect request can not be handled by this service");
         }
         PaymentRedirectActivity.clearResultUri();
-        ChromeCustomTabs.open(context, RedirectUriBuilder.createUri(redirect));
+        Uri uri = null;
```
hjhill, line 58: Unnecessary pre-initialization with `null` - this might hide future bugs.

payment/src/main/java/net/optile/payment/ui/redirect/RedirectUriBuilder.java
```diff
@@ -20,12 +21,12 @@
...
     /**
-     * Construct a Uri from the provided redirect
+     * Create a Uri from the provided redirect
      *
      * @param redirect to be converted into a Uri
      * @return the created Uri
      */
-    public static Uri createUri(Redirect redirect) {
+    public static Uri fromRedirect(Redirect redirect) {
         Uri uri = Uri.parse(redirect.getUrl().toString());
```
hjhill, line 30: Would
```suggestion
        Uri uri = fromURL(redirect.getUrl());
```
be better to avoid code duplication and later inconsistencies?

payment/src/test/java/net/optile/payment/ui/redirect/RedirectRequestTest.java
```diff
@@ -0,0 +1,55 @@
...
+        RedirectRequest.fromOperationResult(operationResult);
+    }
+
+    @Test
+    public void fromOperationResult_success() throws PaymentException, MalformedURLException {
+        OperationResult operationResult = new OperationResult();
+        operationResult.setRedirect(new Redirect());
+        operationResult.setLinks(createLinks());
+        RedirectRequest.fromOperationResult(operationResult);
+    }
```
hjhill, line 47: This test is quite weak - it should actually test something (i.e., have an `assert...`).

payment/src/test/java/net/optile/payment/ui/redirect/RedirectRequestTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+    private Map<String, URL> createLinks() throws MalformedURLException {
```
hjhill, line 49: Two things here:
1. If the Map is only read, you could use `Collections.singletonMap` to create it, making the code more compact and easier to read.
2. Just an idea: I would think about catching `MalformedURLException` and transform it to some sort of `RuntimeException` (for example `IllegalStateException`) - this exception should actually never happen, but its declaration clutters all of the clients of this method. Maybe it would even make sense to create a small `UrlFactory` class for that purpose.

payment/src/main/java/net/optile/payment/ui/redirect/RedirectRequest.java
siebeprojects: Yep oops, fixed this.

payment/src/main/java/net/optile/payment/ui/redirect/RedirectUriBuilder.java
siebeprojects: The fromRedirect() method is now using the fromURL method.

payment/src/test/java/net/optile/payment/ui/redirect/RedirectRequestTest.java
siebeprojects: Added an assertion that the created request, link and redirect are not null.

payment/src/test/java/net/optile/payment/ui/redirect/RedirectRequestTest.java
siebeprojects: Added a TestUtils with helper methods to create a custom URL or default test URL. If the URL cannot be created it will throw the IllegalStateException in order to avoid the exception declarations in the tests.

payment/src/main/java/net/optile/payment/ui/redirect/RedirectService.java
siebeprojects: Removed the null value.

payment/src/test/java/net/optile/test/util/TestUtils.java
```diff
@@ -0,0 +1,43 @@
...
+     *
+     * @param url to be used to create the URL Object
+     * @return the newly created URL
+     * @throws IllegalStateException when the url could not be created
+     */
+    public final static URL createTestURL(String url) {
+        try {
+            return new URL(url);
+        } catch (MalformedURLException e) {
+            throw new IllegalStateException(e);
```
hjhill, line 40: When you extract this function and have the url as a parameter, `IllegalArgumentException` would probably be better. (`IllegalStateException` was meant for the hardcoded URL string that is not actually a parameter, sorry for not being more explicit)

hjhill (approved):

PR #193, opened 2020-12-03
PCX-1208: Fix duplication root warning

siebeprojects:


shared-test/src/main/java/net/optile/sharedtest/service/ListConfig.java
```diff
@@ -0,0 +1,52 @@
...
+
+package net.optile.sharedtest.service;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+/**
+ * Class containing the configuration for creating a new list
+ */
+public final class ListConfig {
```
bzlat: Is this class going to be used by you or by users of the `SDK` as well? If the latter, then some thoughts:

1. Instances of `ListConfig` are the payload of `LIST` payment session requests. So the name feels a little of to me.   Maybe `ListRequest`, `ListData`...?
2. I would add an `Objects.requireNonNull(source, "message that makes sense here")` in the constructor.
The earlier a `null` is announced, the better IMHO.
3. The way instances are used, feels a lot like a builder:
```
ListRequest.of(jsonSource).withAmount(...).withLanguage(...).withPresetFirst(...).withCallbackAppId(...).build();
```
or if you don't like the `withXxx` prefix, then:
```
ListRequest.of(jsonSource).amount(...).language(...).presetFirst(...).callbackAppId(...).build();
```
feels more natural to me.

Feel free to disagree.

bzlat (approved):

shared-test/src/main/java/net/optile/sharedtest/service/ListConfig.java
siebeprojects: Changed to ListRequest and added a builder to it to create a ListRequest Object from the settings.

bzlat (approved):

PR #194, opened 2020-12-09
PCX-1232: Fix focus problems with input fields

siebeprojects:


payment/src/main/java/net/optile/payment/ui/widget/InputLayoutWidget.java
```diff
@@ -90,7 +94,6 @@ public boolean setLastImeOptionsWidget() {
     @Override
     public boolean requestFocus() {
         if (textInput.requestFocus()) {
-            presenter.showKeyboard();
             return true;
         }
         return false;
```
hjhill, line 99: Simpler:

```suggestion
        return textInput.requestFocus();
```

siebeprojects: Oh my, yes that I should have seen myself, thanks.

hjhill (approved):

PR #200, opened 2021-01-07
PCX-1261: Refactor and add model classes

siebeprojects:


payment/src/main/java/net/optile/payment/model/AccountFormData.java
```diff
@@ -8,29 +8,16 @@
...
 package net.optile.payment.model;
 
+import lombok.Getter;
+import lombok.Setter;
+
 /**
  * Account data what should be used to pre-fill payment form.
  */
+@Getter
+@Setter
```
tzie0062, line 18: `@Data` will give you getters, setters, equals&hashcode + toString

bzlat (approved):

PR #201, opened 2021-01-12
PCX-1189: Send UserAgent for requests

siebeprojects:


payment/src/main/java/net/optile/payment/network/UserAgentBuilder.java
```diff
@@ -0,0 +1,133 @@
...
+                setAppVersionCode(packageInfo.versionCode).
+                setAppPackageName(appPackageName).
+                setAppName(packageManager.getApplicationLabel(appInfo).toString()).
+                setBuildManufacturer(android.os.Build.MANUFACTURER).
+                setBuildModel(android.os.Build.MODEL).
+                setBuildVersionSdkInt(android.os.Build.VERSION.SDK_INT).
+                setBuildVersionRelease(android.os.Build.VERSION.RELEASE).
+                build();
+        } catch (final PackageManager.NameNotFoundException e) {
+            throw new PaymentException(e);
```
bzlat, line 80: Interrupting payment flow, if the header cannot be built?

payment/src/main/java/net/optile/payment/network/UserAgentBuilder.java
```diff
@@ -0,0 +1,133 @@
...
+            "App/" + appVersionName + " (" +
+            appPackageName + "; " +
+            appName + "; " +
+            appVersionCode + ") " +
+
+            "Platform/" + buildVersionSdkInt + " (" +
+            buildManufacturer + "; " +
+            buildModel + "; " +
+            buildVersionRelease + ")";
+    }
```
bzlat, line 48: Technically, all properties could be `null`, which would result in a header value like:
```
android-sdk/null (null) App/null (null; null; null) Platform/null (null; null; null)
```

On the other hand, [line 80](https://github.com/optile/android-sdk/pull/201/files#diff-aef50e3d649ef8d9dea39cd735445d79157ee6439fb7e1aac5e716d53870dce1R80) breaks the payment flow with exception, if `PackageManager.NameNotFound` is thrown. Seems a bit inconsistent

payment/src/test/java/net/optile/payment/network/UserAgentBuilderTest.java
```diff
@@ -0,0 +1,37 @@
...
+        UserAgentBuilder.createFromContext(null);
+    }
+
+    @Test
+    public void getValue_succes() throws PaymentException {
+        Context context = ApplicationProvider.getApplicationContext();
+        String value = UserAgentBuilder.createFromContext(context);
+        assertTrue(value.startsWith("android-sdk/"));
+    }
+}
```
bzlat, line 37: Is the header value format still under discussion? 
If it is finalised, shouldn't there be a test to assert the expected format is built?

payment/src/main/java/net/optile/payment/network/BaseConnection.java
```diff
@@ -58,50 +58,38 @@
 
     /**
      * Construct a new BaseConnection
+     *
+     * @param context used to initialize the custom UserAgent
      */
-    BaseConnection() {
-
+    BaseConnection(Context context) {
+        Objects.requireNonNull(context);
```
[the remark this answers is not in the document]
siebeprojects: Replaced the Objects.requireNonNull with IllegalArgumentException which is commonly used in the Android-SDK.

payment/src/main/java/net/optile/payment/network/UserAgentBuilder.java
siebeprojects: Removed the PaymentException, the NameNotFoundException will be logged and a null value will be returned if the useragent could not be created.

payment/src/main/java/net/optile/payment/network/UserAgentBuilder.java
siebeprojects: Added a wrapper around adding string values, so null will not be printed but instead an "" empty string will be printed.

payment/src/test/java/net/optile/payment/network/UserAgentBuilderTest.java
siebeprojects: Yes, there will be a follow up ticket to handle optional values. This certainly could affect the information stored in the user-agent value. However, an extra test has been added that tests all functions of the builder class and that it results in a properly formatted userAgent header value.

bzlat (approved):

PR #202, opened 2021-01-20
PCX-1263: BrowserData for 3DS2

siebeprojects:


payment/src/main/java/net/optile/payment/form/Operation.java
```diff
@@ -96,25 +111,43 @@ public void putValue(String name, Object value) throws PaymentException {
...
-                    break;
-                case PaymentInputType.ALLOW_RECURRENCE:
-                case PaymentInputType.AUTO_REGISTRATION:
-                    form.put(name, value);
-            }
-        } catch (JSONException e) {
-            String msg = "Operation.putValue failed for name: " + name;
-            throw new PaymentException(msg, e);
+        AccountInputData account = operationData.getAccount();
+        String strValue = value.toString();
```
bzlat, line 115: You have a check for `name`, but if `value` is `null` you have a `NPE` here. If the `@NotNull` for `value` is guaranteed elsewhere, please ignore this comment

bzlat (approved):

bzlat (approved):
```
value != null ? value.toString() : null
```
can be replaced by this: [Objects.toString(value, null)](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#toString-java.lang.Object-java.lang.String-)

PR #203, opened 2021-01-25
PCX-1277: Extend support for AccountInputData

siebeprojects:


payment/src/main/java/net/optile/payment/form/Operation.java
```diff
@@ -100,53 +100,101 @@ public void writeToParcel(Parcel out, int flags) {
...
         if (TextUtils.isEmpty(name)) {
             throw new IllegalArgumentException("Name cannot be null or empty");
         }
         AccountInputData account = operationData.getAccount();
-        String strValue = value != null ? value.toString() : null;
+        switch (name) {
+            case PaymentInputType.OPTIN:
+                account.setOptIn(value);
+                break;
+            case PaymentInputType.ALLOW_RECURRENCE:
```
bzlat, line 119: **Suggestion:**
`PaymentInputType` could be converted to an `Enum` and then made into a strategy with double dispatch and instead of these massive `switch` statements, the code could become:
```
final PaymentInputType pit = PaymentInputType.valueOf(name);
pit.setValue(value, operationData);
```

bzlat (approved):

PR #206, opened 2021-02-10
PCX-1209: Reduce warning and cleanup code

siebeprojects:


payment/src/main/java/com/payoneer/mrs/payment/network/ListConnection.java
```diff
@@ -115,12 +113,10 @@ public ListResult getListResult(final String url) throws PaymentException {
...
-                    return handleGetListResultOk(readFromInputStream(conn));
-                default:
-                    throw createPaymentException(rc, conn);
+            if (rc == HttpURLConnection.HTTP_OK) {
+                return handleGetListResultOk(readFromInputStream(conn));
             }
+            throw createPaymentException(rc, conn);
         } catch (JsonParseException | MalformedURLException | SecurityException e) {
             throw createPaymentException(e, false);
         } catch (IOException e) {
```
bzlat, line 122: Why leave this one alone? You handle it the same way as those above.

payment/src/main/java/com/payoneer/mrs/payment/network/UserAgentBuilder.java
```diff
@@ -65,11 +66,17 @@ static String createFromContext(Context context) {
             ApplicationInfo appInfo = packageManager.getApplicationInfo(appPackageName, 0);
             PackageInfo packageInfo = packageManager.getPackageInfo(appPackageName, 0);
 
+            int versionCode;
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+                versionCode = (int) packageInfo.getLongVersionCode();
+            } else {
+                versionCode = packageInfo.versionCode;
+            }
```
bzlat, line 74: **Nitpick:** I have a thing with `if-else`, so my suggestions are:
```
final int versionCode = (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P)
    ? (int) packageInfo.getLongVersionCode()
    : packageInfo.versionCode;
```
or even
```
final int versionCode = (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) ? (int) packageInfo.getLongVersionCode() : packageInfo.versionCode;
```

You can absolutely ignore this comment

payment/src/main/java/com/payoneer/mrs/payment/resource/PaymentGroup.java
```diff
@@ -27,14 +28,16 @@ private PaymentGroup() {
      * @return the id of this PaymentGroup
      */
     public String getId() {
-        return items.get(0).getCode();
+        return items != null ? items.get(0).getCode() : null;
```
bzlat, line 31: Two things:

1. You check for `items != null` or `items == null` everywhere. Why not enforce a `non-null` already in the `setter`?
2. Both this
```
return items != null ? items.get(0).getCode() : null;
``` 
and this
```
if (item.getCode().equals(code)) {
    return item.getRegex();
}
```
are vulnerable to `IndexOutOfBoundsException` and/or `NullPointerException`, because `items != null` is not a guarantee for `items.get(0) != null`.

payment/src/main/java/com/payoneer/mrs/payment/resource/PaymentGroup.java
```diff
@@ -47,7 +50,9 @@ public void populate(Map<String, PaymentGroup> map) {
      * @return the regex for smart selection or null if not found
      */
     public String getSmartSelectionRegex(String code) {
-
+        if (items == null) {
+            return null;
+        }
         for (PaymentGroupItem item : items) {
 
             if (item.getCode().equals(code)) {
```
bzlat, line 58: Two things:
1. For the reasons given above, regarding `IOBE` and `NPE`, a better alternative is:
```
Objects.equals(item.getCode(), code)
```
with an obligatory `item != null` check, if you'd like to be covered in all cases.

payment/src/main/java/com/payoneer/mrs/payment/resource/ValidationGroup.java
```diff
@@ -46,7 +51,9 @@ public String getValidationRegex(String type) {
     }
 
     public ValidationGroupItem getGroupItem(String type) {
-
+        if (items == null) {
+            return null;
+        }
         for (ValidationGroupItem item : items) {
             if (item.getType().equals(type)) {
```
bzlat, line 58: Same applies here when defending from `IOBE` and/or `NPE`

payment/src/main/java/com/payoneer/mrs/payment/ui/page/BasePaymentActivity.java
```diff
@@ -182,6 +169,21 @@ public void close() {
...
     }
 
+    /**
+     * Show Hint dialog
+     *
+     * @param networkCode Code if the network
+     * @param type type if input field
+     * @param listener listening to events of this Hint Dialog.
+     */
+    public void showHintDialog(String networkCode, String type, PaymentDialogListener listener) {
```
bzlat, line 179: This code:
```
if (active) {
    PaymentDialogFragment dialog = PaymentDialogHelper.createHintDialog(networkCode, type, listener);
    showPaymentDialog(dialog);
}
```
has the following advantages:
1. The `if` condition reads better and there's no danger of not seeing the negation.
2. The completely unnecessary `return` is gone

bzlat (approved):

payment/src/main/java/com/payoneer/mrs/payment/resource/PaymentGroup.java
siebeprojects: I merged too quickly, yes I will fix this in a follow up PR.

payment/src/main/java/com/payoneer/mrs/payment/network/UserAgentBuilder.java
siebeprojects: The Build.VERSION_SDK_INT check has to be done in an if statement, else the compiler will still complain.

payment/src/main/java/com/payoneer/mrs/payment/resource/PaymentGroup.java
siebeprojects: These classes are normally constructed from a JSON File and thus I cannot guarantee that "items" will always be non-null unless we will start adding annotations.

PR #207, opened 2021-02-11
PCX-1209: minor updated to reduce warnings

siebeprojects:


payment/src/main/java/com/payoneer/mrs/payment/resource/PaymentGroup.java
```diff
@@ -55,7 +56,7 @@ public String getSmartSelectionRegex(String code) {
         }
         for (PaymentGroupItem item : items) {
 
-            if (item.getCode().equals(code)) {
+            if (Objects.equals(item.getCode(), code)) {
```
bzlat, line 59: This one is trickier - now if `item.getCode() == null == code` you will return `item.getRegex()`. Unless `item.getCode()` is never `null`

siebeprojects: Yes, I presume that item.getCode() will never be null.

bzlat (approved):

PR #209, opened 2021-02-18
PCX-1275: Support optional user-agent sections

siebeprojects:


siebeprojects:
This ticket can be tested with the BrowserStack app: Example-SDK-PCX-1275-debug.apk

PR #234, opened 2021-05-31
PCX-1659: Fix listresult.json for tests

siebeprojects:
Link logo URLs to Github logos in checkout-android repo.
Add listresult_logoerror.json for testing logo error handling.


Sorix (changes requested):
Please don't remove any properties from JSON because some candidates have written their own models and that will fail JSON parsing.

Using our models wasn't a requirement, that was a recommendation, so they don't know what fields are mandatory and what are not.

Sorix (approved):
Now it works, thank you!

PR #236, opened 2021-06-08
PCX-1384: Add delete saved account functionality

siebeprojects:


checkout/src/main/java/com/payoneer/checkout/form/DeleteAccount.java
```diff
@@ -0,0 +1,80 @@
...
+        this.url = (URL) in.readSerializable();
+
+        try {
+            GsonHelper gson = GsonHelper.getInstance();
+            deregistrationData = gson.fromJson(in.readString(), DeregistrationData.class);
+        } catch (JsonSyntaxException e) {
+            // this should never happen since we use the same GsonHelper
+            // to produce these Json strings
+            Log.w("Checkout", e);
+            throw new RuntimeException(e);
```
bzlat, line 55: Logging and throwing at the same time is unnecessary

checkout/src/main/java/com/payoneer/checkout/network/BaseConnection.java
```diff
@@ -124,6 +125,21 @@ HttpURLConnection createGetConnection(final URL url) throws IOException {
...
     }
 
+    /**
+     * Creates a new HTTP DELETE connection
+     *
+     * @param url the Url pointing to the Payment API
+     * @return HttpURLConnection a HttpURLConnection object
+     */
+    HttpURLConnection createDeleteConnection(final URL url) throws IOException {
+        final HttpURLConnection conn = (HttpURLConnection) url.openConnection();
```
bzlat, line 135: Do you ever want to have non-HTTPS connections?

checkout/src/main/java/com/payoneer/checkout/redirect/RedirectRequest.java
```diff
@@ -22,18 +22,30 @@
...
+        this.requestType = requestType;
         this.redirect = redirect;
         this.link = link;
     }
 
+    /**
+     * Get the type of this request
+     *
+     * @return the type of this request
+     */
```
bzlat, line 44: A couple of suggestions:

1. This `javadoc` is not very helpful. The `Get the type of this request` is already covered by the `getRequestType` method name. It would be nice, if there's a bit more details as to what that type might be - is this about the `GET vs POST` or something else?
2. I find the use of `int` here a code smell. If I understood the intention of this `field` correctly, it has to do with the flags defined in the `BaseNetworkService`:
```
    private final static int PROCESS_PAYMENT = 0;
    private final static int DELETE_ACCOUNT = 1;
```
Using `enum RequestType` type to express this concept feels much better to me.

checkout/src/main/java/com/payoneer/checkout/ui/list/CardEventHandler.java
```diff
@@ -0,0 +1,135 @@
...
+import com.payoneer.checkout.ui.widget.FormWidget;
+import com.payoneer.checkout.ui.widget.WidgetPresenter;
+import com.payoneer.checkout.validation.ValidationResult;
+import com.payoneer.checkout.validation.Validator;
+
+import android.view.View;
+
+/**
+ * Class handling events originating from payment cards and widgets
+ */
```
bzlat, line 22: I'm a bit confused here, so please bear with me: it seems to me that the `CardEventHandler` is actually delegating to a `PaymentCardListener`. My mental model is:
```
Event Source ---- sends events to ----> Event Listener(s) ---- delegates to ----> Event Handler(s)
```
What am I missing here?

checkout/src/main/java/com/payoneer/checkout/ui/list/IconView.java
```diff
@@ -0,0 +1,52 @@
...
+class IconView {
+    private final ViewSwitcher switcher;
+
+    /**
+     * Construct a new IconView
+     *
+     * @param parent of this view
+     */
+    IconView(View parent) {
+        switcher = parent.findViewById(R.id.viewswitcher_icon);
```
bzlat, line 28: This is just out of curiosity: what guarantees that `findViewById` will always return a `non-null` switcher?

checkout/src/main/java/com/payoneer/checkout/form/DeleteAccount.java
siebeprojects: Thats correct, removed these logs.

checkout/src/main/java/com/payoneer/checkout/network/BaseConnection.java
siebeprojects: I believe the reason for using the HttpURLConnection was that for different versions of Android the package of the HttpsURLConnection class was different. So, casting would fail on some Android devices and it was necessary to cast to the HttpURLConnection.

checkout/src/main/java/com/payoneer/checkout/redirect/RedirectRequest.java
siebeprojects: Because of performance and memory, Enums are avoided in Android. 
The use of integer values to identify results or requests is heavily used within Android as well.
I updated the Javadocs and changed the name from requestType to requestCode as this is used as well in other parts of the Checkout SDK.

checkout/src/main/java/com/payoneer/checkout/ui/list/CardEventHandler.java
siebeprojects: Updated the Javadocs, hopefully it is clearer now.

checkout/src/main/java/com/payoneer/checkout/ui/list/IconView.java
siebeprojects: It does not guarantee it, the compiler will tell if the R.id.viewswitcher_icon exists but this would still not guarantee that the parent View would have this View element bound to this ID. 
The question is then, should we throw a NullPointerException by user e.g. Objects.requireNoNull(switcher)?

bzlat: Well, the `NPE` will be out and about anyway, so unless there's a nice way to handle it, it doesn't make much difference. I was just interested in what's the recommended way of handling this.

checkout/src/main/java/com/payoneer/checkout/redirect/RedirectRequest.java
bzlat: Interesting, good to know. Maybe it's worth looking into the `support-annotations`. They solve this problem nicely:

```
final static int PROCESS_PAYMENT = 0;
final static int DELETE_ACCOUNT = 1;

@IntDef({PROCESS_PAYMENT, DELETE_ACCOUNT})
@Retention(RetentionPolicy.SOURCE)
public @interface RequestType {}

public static RedirectRequest fromOperationResult(@RequestType int requestType, ... 
```

bzlat (approved):

PR #244, opened 2021-07-08
PCX-1776: Create automated UI tests for registration of networks

siebeprojects:


checkout/src/main/java/com/payoneer/checkout/ui/page/PaymentListActivity.java
```diff
@@ -169,14 +165,19 @@ public IdlingResource getLoadIdlingResource() {
...
     }
 
+    /**
+     * Only called from test, reset the load idling resource
+     */
+    @VisibleForTesting
+    public void resetLoadIdlingResource() {
+        loadIdlingState = false;
+        setIdlingResourceState(loadIdlingResource, false);
+    }
```
konurbaev-e, line 179: What do you think about creating a test class in the same package and call `setLoadIdlingState(false)` from it?
In that case, you won't need this public method in the main sources.

siebeprojects: It has to be a public method, this method is called from the automated UI tests themselves. 
With Espresso test environment from Android, the tests themselves have a reference to the PaymentListActivity and can only call functions inside the activity. So, in Espresso, it is so that tests communicate with the sources through IdlingResources.

siebeprojects: @konurbaev-e the creation, setting and resetting of IldingResources have now been moved to the PaymentIdlingResources class. Automated UI tests obtain a reference to this class through the Activity and can then obtain the required IldingResource from that class.

bzlat (approved):

PR #246, opened 2021-08-12
PCX-1382: Update flow

siebeprojects:
This feature allow users to create, modify and delete saved accounts.


checkout/src/main/java/com/payoneer/checkout/ui/list/PaymentItemList.java
```diff
@@ -0,0 +1,70 @@
...
+    int getItemCount() {
+        return items.size();
+    }
+
+    int getItemViewType(int index) {
+        return items.get(index).viewType;
+    }
+
+    void addItem(ListItem item, boolean preselected) {
+        items.add(item);
```
bzlat, line 65: Is there something else that prevents `ListItem item` from being `null`? If items are added to the list without that `null` check, line `49` above is a `NPE`.

bzlat (approved):

PR #251, opened 2021-08-19
PCX-1780: Add 3DS2 functional tests using TESTPSP adapter

siebeprojects:


example-checkout/src/androidTest/java/com/payoneer/checkout/examplecheckout/PayPalRedirectTests.java
```diff
@@ -46,7 +46,9 @@ public void testPayPalRedirect_directCharge_browserClosed() {
         PaymentListHelper.openPaymentListCard(networkCardIndex, "card_network");
         PaymentListHelper.clickPaymentListCardButton(networkCardIndex);
 
-        clickDecisionPageButton("com.android.chrome:id/close_button");
+        clickCustomerDecisionPageButton("com.android.chrome:id/close_button");
+        UiDeviceHelper.waitUiObjectHasPackage("com.payoneer.checkout.examplecheckout");
```
bzlat, line 50: This is the same as `waitForAppRelaunch()` which you inherit from `AbstractTest`. Why not use it here?

example-checkout/src/androidTest/java/com/payoneer/checkout/examplecheckout/Three3DS2Tests.java
```diff
@@ -63,8 +69,38 @@ public void testVisa_directCharge_PROCEED_PENDING() {
         PaymentListHelper.fillPaymentListCard(groupCardIndex, TestDataProvider.visaCardTestData());
         PaymentListHelper.clickPaymentListCardButton(groupCardIndex);
 
+        clickDeviceCollectionPagePageButton("customer-abort");
+        clickCustomerDecisionPageButton("customer-accept");
+        waitForAppRelaunch();
+
         register(resultIdlingResource);
-        matchResultInteraction(InteractionCode.PROCEED, InteractionReason.PENDING);
+        matchResultInteraction(InteractionCode.PROCEED, InteractionReason.OK);
```
bzlat, line 77: Two things:

1. I think this should test `InteractionReason.OK || InteractionReason.PENDING`. The `3DS2` flow may be both sync or async, depending on provider/adapter and this is could be changed at any moment in the `TestPSP`. Both are valid and I don't think the test here cares if the flow is sync or async, does it?
2. The method name is `testXxx_PROCEED_PENDING` and the internals now test `PROCEED_OK`.

example-checkout/src/androidTest/java/com/payoneer/checkout/examplecheckout/PayPalRedirectTests.java
siebeprojects: Yes, it should be also used here. Will change it.

example-checkout/src/androidTest/java/com/payoneer/checkout/examplecheckout/Three3DS2Tests.java
siebeprojects: Changed the method names and kept the check for the OK reason.

PR #260, opened 2021-09-13
PCX-2020: Combine autoRegistration and allowRecurrence in one Checkbox

siebeprojects:


checkout/src/main/java/com/payoneer/checkout/localization/LocalLocalizationHolder.java
```diff
@@ -56,5 +57,7 @@ public LocalLocalizationHolder(Context context) {
 
         map.put(MESSAGES_UNSAVED_TITLE, context.getString(R.string.messages_unsaved_title));
         map.put(MESSAGES_UNSAVED_TEXT, context.getString(R.string.messages_unsaved_text));
+
```
bzlat, line 60: I know you don't like extra spaces. I wonder what's the purpose of this one here? (Not sarcasm, just genuine curiosity)

checkout/src/main/java/com/payoneer/checkout/ui/service/RegistrationOptionsBuilder.java
```diff
@@ -30,95 +38,87 @@
...
     }
 
-    private RegistrationOption buildUpdateAllowRecurrenceOption() {
-        switch (allowRecurrence) {
-            case RegistrationType.NONE:
-                return new RegistrationOption(CheckboxMode.NONE, ALLOW_RECURRENCE_FORCED);
-            default:
-                return new RegistrationOption(CheckboxMode.FORCED, ALLOW_RECURRENCE_FORCED);
+    private int getRegistrationOptionIndex() throws PaymentException {
+        for (int i = 0, e = VALID_COMBINATIONS.length; i < e; i++) {
```
bzlat, line 103: This is interesting too. Does it have any benefit?

checkout/src/main/java/com/payoneer/checkout/ui/service/RegistrationOptionsBuilder.java
```diff
@@ -30,95 +38,87 @@
...
+        { NONE, NONE, CheckboxMode.NONE },
+        { FORCED, NONE, CheckboxMode.NONE },
+        { FORCED, NONE, CheckboxMode.NONE },
+        { FORCED, FORCED, CheckboxMode.NONE },
+        { FORCED, FORCED, CheckboxMode.NONE },
+        { FORCED, NONE, CheckboxMode.NONE },
+        { FORCED, NONE, CheckboxMode.NONE },
+        { FORCED, FORCED, CheckboxMode.NONE },
+        { FORCED, FORCED, CheckboxMode.NONE }
+    };
```
bzlat, line 75: Are the repeated values here a cheat, so that `DEFAULT_VALUES` and `UPDATE_VALUES` have the same number of elements and the `getRegistrationOptionIndex()` work, or am I missing the point?

checkout/src/main/java/com/payoneer/checkout/ui/service/RegistrationOptionsBuilder.java
siebeprojects: Actually here it has no purpose, I still do it for method calls because they are expensive and I dont like to rely on the JVM. I will change this.

checkout/src/main/java/com/payoneer/checkout/ui/service/RegistrationOptionsBuilder.java
siebeprojects: For both flows (default and update) we need to first check if a valid combination is provided, I use the index for that. For the second step (create the RegistrationOptions) I use the same index to obtain the values for the specific flow and combination. I don't see it as a cheat.

checkout/src/main/java/com/payoneer/checkout/localization/LocalLocalizationHolder.java
siebeprojects: I have spaces between groups of local localisations, just to make it a bit easiest to see which localizations I have per section.

bzlat (approved):

bzlat (approved):

PR #266, opened 2021-10-07
PCX-2179: remove glide

Fbada006:
Remove all implementations of glide and use a custom one.

### DIFF
--- a/checkout/src/main/java/com/payoneer/checkout/network/ImageConnection.java
+++ b/checkout/src/main/java/com/payoneer/checkout/network/ImageConnection.java
@@ -1,0 +1,53 @@
+/*
+ *   Copyright (c) 2021 Payoneer Germany GmbH
+ *   https://www.payoneer.com
+ *
+ *   This file is open source and available under the MIT license.
+ *   See the LICENSE file for more information.
+ */
+
+package com.payoneer.checkout.network;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+import com.payoneer.checkout.core.PaymentException;
+
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
+
+/**
+ * Class for loading images from the network
+ * All requests in this class are blocking calls and should be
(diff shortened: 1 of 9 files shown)


checkout/src/main/java/com/payoneer/checkout/network/ImageConnection.java
```diff
@@ -0,0 +1,55 @@
+/*
+ *
+ *  * Copyright (c) 2021 Payoneer Germany GmbH
```
siebeprojects, line 3: This is not correct, there are double quotes * *.

siebeprojects (commented):
The format of the copyright statement is wrong and we should only apply it for now to Java files.

checkout/src/main/java/com/payoneer/checkout/util/NetworkLogoLoader.java
```diff
@@ -50,20 +65,59 @@ public static NetworkLogoLoader getInstance() {
...
     }
 
-    private void loadIntoView(ImageView view, String networkCode, URL networkLogoUrl) {
+    private void loadImageIntoView(ImageView view, String networkCode, URL networkLogoUrl) {
+        final Context context = view.getContext();
         if (localNetworkLogos.size() == 0) {
             loadLocalNetworkLogos(view.getContext());
         }
-        String url;
+        if (viewHasImage(view)) {
```
siebeprojects, line 76: This "if" statement should be removed as it fails showing the correct logo when using smart switch.

checkout/src/main/java/com/payoneer/checkout/util/NetworkLogoLoader.java
```diff
@@ -81,6 +135,16 @@ private void loadLocalNetworkLogos(Context context) {
         }
     }
 
+    private boolean viewHasImage(ImageView view) {
```
siebeprojects, line 138: This method is not required anymore since we remove the optimisation that checks if an ImageView already has a Drawable set.

siebeprojects (commented):
We should remove the optimisation checking if the ImageView already has a Drawable.

PR #275, opened 2021-11-10
PCX-2263: Remove required preset account object when charging preset account

Fbada006:
More details in the ticket [here](https://optile.atlassian.net/browse/PCX-2263)

### DIFF
--- a/checkout/src/main/java/com/payoneer/checkout/ui/page/ChargePaymentActivity.java
+++ b/checkout/src/main/java/com/payoneer/checkout/ui/page/ChargePaymentActivity.java
@@ -30,7 +30,7 @@ public final class ChargePaymentActivity extends BasePaymentActivity implements 
     private final static String EXTRA_OPERATION_TYPE = "operation_type";
     public final static int TYPE_CHARGE_OPERATION = 1;
     public final static int TYPE_CHARGE_PRESET_ACCOUNT = 2;
-    private int OPERATION_TYPE;
+    private int chargeType;
     private ChargePaymentPresenter presenter;
     private Operation operation;
 
@@ -90,7 +90,7 @@ public final class ChargePaymentActivity extends BasePaymentActivity implements 
         Bundle bundle = savedInstanceState == null ? getIntent().getExtras() : savedInstanceState;
         if (bundle != null) {
             this.operation = bundle.getParcelable(EXTRA_OPERATION);
-            this.OPERATION_TYPE = bundle.getInt(EXTRA_OPERATION_TYPE);
+            this.chargeType = bundle.getInt(EXTRA_OPERATION_TYPE);
         }
         setContentView(R.layout.activity_chargepayment);
 
@@ -124,7 +124,7 @@ public final class ChargePaymentActivity extends BasePaymentActivity implements 
     @Override
     public void onResume() {
         super.onResume();
-        presenter.onStart(operation, OPERATION_TYPE);
+        presenter.onStart(operation, chargeType);
     }
 
     /**
--- a/checkout/src/main/java/com/payoneer/checkout/ui/page/ChargePaymentActivity.java
+++ b/checkout/src/main/java/com/payoneer/checkout/ui/page/ChargePaymentActivity.java
@@ -28,8 +28,8 @@ public final class ChargePaymentActivity extends BasePaymentActivity implements 
 
     private final static String EXTRA_OPERATION = "operation";
     private final static String EXTRA_OPERATION_TYPE = "operation_type";
-    private final static int TYPE_CHARGE_OPERATION = 1;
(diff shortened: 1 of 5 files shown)


checkout/src/main/java/com/payoneer/checkout/ui/page/ChargePaymentActivity.java
```diff
@@ -31,6 +27,10 @@
 public final class ChargePaymentActivity extends BasePaymentActivity implements BasePaymentView {
 
     private final static String EXTRA_OPERATION = "operation";
+    private final static String EXTRA_OPERATION_TYPE = "operation_type";
+    public final static int TYPE_CHARGE_OPERATION = 1;
+    public final static int TYPE_CHARGE_PRESET_ACCOUNT = 2;
+    private int OPERATION_TYPE;
```
siebeprojects, line 33: OPERATION_TYPE should be written in lower-case since it is a normal variable, perhaps "chargeType" be a better name.

checkout/src/main/java/com/payoneer/checkout/ui/page/ChargePaymentPresenter.java
```diff
@@ -83,7 +91,10 @@ void onStop() {
     public void onPaymentSessionSuccess(PaymentSession session) {
         ListResult listResult = session.getListResult();
         Interaction interaction = listResult.getInteraction();
-
+        if (chargeType == ChargePaymentActivity.TYPE_CHARGE_PRESET_ACCOUNT) {
```
siebeprojects, line 94: This code blocks should go into the handleLoadSessionProceed() method. We only should process a presetAccount when the Interaction code of the list is PROCEED.

The handleLoadSessionProceed method could look like this.

```
private void handleLoadSessionProceed(PaymentSession session) {
        if (chargeType == TYPE_CHARGE_PRESET_ACCOUNT) {
            PresetAccount account = session.getPresetAccount();
            if (account == null) {
                closeWithErrorCode("PresetAccount not found in ListResult");
                return;
            }
            this.operation = Operation.fromPresetAccount(account);
        }
        else if (!session.containsOperationLink(operation.getURL())) {
            closeWithErrorCode("Operation not found in ListResult");
            return;
        }
        try {
            this.session = session;
            networkService = loadNetworkService(operation.getNetworkCode(), operation.getPaymentMethod());
            networkService.setListener(this);
            processPayment(operation);
        } catch (PaymentException e) {
            closeWithErrorCode(PaymentResultHelper.fromThrowable(e));
        }
    }

```

siebeprojects (commented):
Small changes where we obtain the PresetAccount and create the Operation from it.

checkout/src/main/java/com/payoneer/checkout/ui/page/ChargePaymentPresenter.java
```diff
@@ -253,4 +264,14 @@ private void loadPaymentSession() {
         view.showProgress(true);
         sessionService.loadPaymentSession(listUrl, view.getActivity());
     }
+
+    public void createOperationForChargePreset(PresetAccount account) {
```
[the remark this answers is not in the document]
Fbada006: So where would we use the operation object that has been created?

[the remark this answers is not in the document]
Fbada006: It should actually be private and it does not need to be static. It will only be used in the presenter

checkout/src/main/java/com/payoneer/checkout/ui/page/ChargePaymentActivity.java
```diff
@@ -31,6 +27,10 @@
 public final class ChargePaymentActivity extends BasePaymentActivity implements BasePaymentView {
 
     private final static String EXTRA_OPERATION = "operation";
+    private final static String EXTRA_OPERATION_TYPE = "operation_type";
```
siebeprojects, line 30: This should be: EXTRA_CHARGE_TYPE

checkout/src/main/java/com/payoneer/checkout/ui/page/ChargePaymentPresenter.java
```diff
@@ -59,8 +62,12 @@
         sessionService.setListener(this);
     }
 
-    void onStart(Operation operation) {
-        this.operation = operation;
+    void onStart(Operation operation, int chargeType) {
+        this.chargeType = chargeType;
+        if (chargeType == ChargePaymentActivity.TYPE_CHARGE_OPERATION) {
+            // The operation object is available. If not, then we need to create the operation object in line 92
```
siebeprojects, line 68: Please do not reference line numbers in comments, these will be invalid once we add/remove code in new stories.

siebeprojects (commented):
Minor name change for the extra field

siebeprojects (approved):

PR #276, opened 2021-11-17
PCX-2289: Scroll and open the correct selected payment card

siebeprojects:
When loading a LIST object with a PRESET account, the first thing viewable in the viewport is the UI card for the PRESET account, while the section title (“Previously selected”) and (potential) warning text are not visible.

### DIFF
--- a/checkout/src/main/java/com/payoneer/checkout/ui/list/ListItem.java
+++ b/checkout/src/main/java/com/payoneer/checkout/ui/list/ListItem.java
@@ -20,12 +20,4 @@ abstract class ListItem {
     ListItem(int viewType) {
         this.viewType = viewType;
     }
-
-    boolean hasPaymentCard() {
-        return false;
-    }
-
-    PaymentCard getPaymentCard() {
-        return null;
-    }
 }
--- a/checkout/src/main/java/com/payoneer/checkout/ui/list/ListItem.java
+++ b/checkout/src/main/java/com/payoneer/checkout/ui/list/ListItem.java
@@ -20,14 +20,4 @@ abstract class ListItem {
     ListItem(int viewType) {
         this.viewType = viewType;
     }
-
-    /**
-     * Check if this ListItem is a HeaderItem
-     *
-     * @param listItem the item to be checked
-     * @return true when a HeaderItem, false otherwise
-     */
-    public final static boolean isHeaderItem(ListItem listItem) {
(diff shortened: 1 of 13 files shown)


checkout/src/main/java/com/payoneer/checkout/ui/list/ListItem.java
```diff
@@ -20,12 +20,4 @@
     ListItem(int viewType) {
         this.viewType = viewType;
     }
-
-    boolean hasPaymentCard() {
-        return false;
-    }
```
Fbada006, line 26: Why are we removing these?

checkout/src/main/java/com/payoneer/checkout/ui/service/PaymentSessionBuilder.java
```diff
@@ -292,4 +293,34 @@ private void addNetwork2GroupCard(Map<String, NetworkCard> cards, PaymentNetwork
...
+     * A card is marked as preselected in the following order:
+     *
+     * 1. The card has a network that is marked selected by the Payment API
+     * 2. The card is presenting a PresetAccount
+     * 3. It is the only card in the entire list
+     *
+     * @param paymentSections containing the sections of PaymentCards
+     */
+    private void setPreselectedCard(List<PaymentSection> paymentSections) {
+        List<PaymentCard> cards = new ArrayList<>();
```
Fbada006, line 308: We can lift the creation of this list outside this method and set it as a field in the class to avoid its creation every time

checkout/src/main/java/com/payoneer/checkout/ui/list/ListItem.java
siebeprojects: It was incorrect to begin with that the ListItem knows about PaymentCards, it has now been replaced by methods that validate if a ListItem is of type HeaderItem or a PaymentCardItem, these methods are then provided by the subclasses themselves.

Fbada006 (approved):

checkout/src/main/java/com/payoneer/checkout/ui/service/PaymentSessionBuilder.java
siebeprojects: True but in this case there is not much of a benefit qua performance. This class is created when the Activity is created and probably the list is max reloaded a few times during a session. We do this for example in the PaymentCard and the userInputTypes array, there we keep the list as a class member and clear it when not used.

PR #295, opened 2022-01-21
PCX-984: adjust credit and debit card masked account display

Fbada006:
Details in ticket


checkout/src/main/java/com/payoneer/checkout/util/PaymentUtils.java
```diff
@@ -117,8 +118,28 @@ public static int toInt(Integer value) {
...
+    public static String getAccountMaskLabel(AccountMask accountMask, String paymentMethod, String networkLabel) {
+        if (isCardPaymentMethod(paymentMethod)) {
+            return getFormattedMaskedDisplay(accountMask.getNumber(), networkLabel);
+        } else if (!TextUtils.isEmpty(accountMask.getIban())) {
+            return getFormattedMaskedDisplay(accountMask.getIban(), networkLabel);
+        }
+        return accountMask.getDisplayLabel();
+    }
+
+    private static String getFormattedMaskedDisplay(String number, String networkLabel) {
```
optile-mobiledev, line 130: Its not a number, it could also be an iban value. The name "Display" could be "Label". 
So perhaps create it to:
createFormattedMaskedLabel(String maskedLabel, String networkLabel) {
}

Fbada006: Done

PR #304, opened 2022-02-21
Feature/PCX-2637

siebeprojects:
**AS A** Checkout merchant
**I WANT** the Android SDK to collect risk-related information from different risk providers and submit it with customer transactions
**SO THAT** the risk for fraudulent transactions is minimised

**Acceptance Criteria**

- Create a empty risk framework for loading different risk provider modules in the Checkout SDK
- Support adding risk providers through loabable modules through the build.gradle file


checkout/src/main/java/com/payoneer/checkout/form/Operation.java
```diff
@@ -119,6 +93,55 @@ public void writeToParcel(Parcel out, int flags) {
...
+    public void setAccountInputData(AccountInputData inputData) {
+        operationData.setAccount(inputData);
+    }
+
+    public void setBrowserData(BrowserData browserData) {
+        operationData.setBrowserData(browserData);
+    }
+
+    public String toJson() {
+        GsonHelper gson = GsonHelper.getInstance();
```
Fbada006, line 141: Instead of creating a `gson` object every time, initialise one at the top of the class as a field then reuse it.

checkout/src/main/java/com/payoneer/checkout/core/WorkerTask.java
```diff
@@ -62,6 +62,30 @@ public boolean isSubscribed() {
         return this.subscriber != null;
     }
 
+    /**
+     * Helper method to unsubscribe the workerTask
+     *
```
Fbada006, line 67: I do not think these comments make the code any more readable. The method names speak for the themselves

checkout/src/main/java/com/payoneer/checkout/form/Operation.java
```diff
@@ -261,56 +306,17 @@ private void putInputElementStringValue(String name, String value) throws Paymen
...
-            builder.append("url=").append(url).append(", ");
-        }
-        if (operationData != null) {
-            builder.append("operationData=").append(operationData);
+    private int getProviderRequestIndex(ProviderParameters request) {
+        List<ProviderParameters> list = operationData.getProviderRequests();
+        if (list != null) {
+            for (int i = 0; i < list.size(); i++) {
+                ProviderParameters parameters = list.get(i);
+                if ((Objects.equals(parameters.getProviderCode(), request.getProviderCode())) &&
```
Fbada006, line 314: Here, couldn't you do something like `parameters.getProviderCode().equals(request.getProviderCode())`?

checkout/src/main/java/com/payoneer/checkout/risk/RiskException.java
```diff
@@ -0,0 +1,36 @@
...
+
+package com.payoneer.checkout.risk;
+
+/**
+ * RiskException containing the details of the risk error that occurred while processing risk providers
+ */
+public class RiskException extends Exception {
+
+    /**
+     * {@inheritDoc}
```
Fbada006, line 17: What are these `inheritedDoc` things? I think we should go ahead and remove them.

checkout/src/main/java/com/payoneer/checkout/risk/RiskProviderController.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Match the code and type of this RiskProviderController with the provided riskProviderCode and riskProviderType.
+     * *
+     * @param riskProviderCode code of the risk provider to match
+     * @param riskProviderType type of the risk provider to match
+     * @return true when the code and type matches, false otherwise
+     */
+    public boolean matches(final String riskProviderCode, final String riskProviderType) {
+        return (Objects.equals(getRiskProviderCode(), riskProviderCode)) &&
```
Fbada006, line 49: Same here for the object equality check

checkout/src/main/java/com/payoneer/checkout/risk/RiskProviderController.java
```diff
@@ -0,0 +1,99 @@
...
+     * @param context contains information about the application environment
+     */
+    public void initialize(final Context context) {
+        String code = info.getRiskProviderCode();
+        String type = info.getRiskProviderType();
+        riskProvider = RiskProviderLookup.getRiskProvider(code, type);
+
+        if (riskProvider == null) {
+            String message = "RiskProviderController(" + code + ", " + type + ") could not find RiskProvider";
+            Log.w("checkout-sdk", message);
```
Fbada006, line 65: One other thing we need to talk about is the logging we do, especially the log tag. We should should class names for that to make it easier to see where they come from. Or we can use something like `ClassName.methodName` as the tag, like the way Timber does it. You pinpoint exactly where the error/message is coming from in case of an issue without wasting too much time

checkout/src/main/java/com/payoneer/checkout/risk/RiskProviderLookup.java
```diff
@@ -0,0 +1,61 @@
...
+     * @param riskProviderType to be used to lookup a RiskProvider
+     * @return the RiskProvider or null if none found
+     */
+    public static RiskProvider getRiskProvider(String riskProviderCode, String riskProviderType) {
+        RiskProviderFactory factory = getRiskProviderFactory(riskProviderCode, riskProviderType);
+        return factory != null ? factory.createRiskProvider() : null;
+    }
+
+    private static RiskProviderFactory getRiskProviderFactory(String providerCode, String providerType) {
+        Objects.requireNonNull(providerCode);
```
Fbada006, line 36: These are supposed to be guards? Why don't we keep using the pattern you have been using of a simple `if/else`

checkout/src/test/java/com/payoneer/checkout/risk/RiskProvidersTest.java
```diff
@@ -0,0 +1,98 @@
...
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
+
+import com.payoneer.checkout.model.ProviderParameters;
+
+import android.content.Context;
+import androidx.test.core.app.ApplicationProvider;
+
+@RunWith(RobolectricTestRunner.class)
+public class RiskProvidersTest {
```
Fbada006, line 27: Love the tests. The more the better

shared-test/lists/listresult_risk.json
```diff
@@ -0,0 +1,316 @@
+{
+    "links": {
+        "self": "https://api.pi-nightly.integration.oscato.com/pci/v1/620bb3cc550f733fb89b30c7lnd26ntqthgm7q3b1avveq1eie",
```
Fbada006, line 3: I cannot seem to find where this is used. Is this a mistake?

checkout/src/main/java/com/payoneer/checkout/form/Operation.java
bzlat: Are you guaranteed that the list always returns `non-null` elements?

checkout/src/main/java/com/payoneer/checkout/risk/RiskException.java
bzlat: See [Javadoc - {@inheritDoc} tag](https://docs.oracle.com/javase/7/docs/technotes/tools/windows/javadoc.html#inheritDoc)

checkout/src/main/java/com/payoneer/checkout/form/Operation.java
Fbada006: I am not sure. Why do you ask?

checkout/src/main/java/com/payoneer/checkout/risk/RiskException.java
Fbada006: Thanks for that.
Still, they do not really add any benefit to the code with regards to readability.

checkout/src/main/java/com/payoneer/checkout/form/Operation.java
siebeprojects: The Gson object is not created each time. It is actually a singleton inside the GsonHelper who maintains the singleton.

checkout/src/main/java/com/payoneer/checkout/form/Operation.java
bzlat: Because if `parameters.getProviderCode() == null`, then `parameters.getProviderCode().equals(...)` will throw `NPE`, while `Objects.equals()` will not.

checkout/src/main/java/com/payoneer/checkout/risk/RiskException.java
bzlat: `Javadoc` is not about code readability, it's about providing a documentation for the users of the SDK, who won't necessarily be looking at the source code all the time.

checkout/src/main/java/com/payoneer/checkout/form/Operation.java
siebeprojects: The providerCode and providerType could be null, e.g. type is optional. 
It is intentional that when both are null, that that ProviderParameters will be replaced at that index.
And it is expected that the list is always returning non-null elements.

checkout/src/main/java/com/payoneer/checkout/risk/RiskProviderController.java
siebeprojects: The code and type could be null and therefor we use the Objects method for checking if they equal.

checkout/src/main/java/com/payoneer/checkout/risk/RiskProviderController.java
siebeprojects: Yes, you are right, we need to refactor these and be more consistent. We can do that in a separate ticket whereby we also refactor other issues like using annotations for null checks, final keywords etc.

checkout/src/main/java/com/payoneer/checkout/risk/RiskProviderLookup.java
siebeprojects: Recently I started using more the Objects.require but maybe the IllegalParameterException is better because it provides better info what went wrong. I will change them in this PR.

checkout/src/main/java/com/payoneer/checkout/form/Operation.java
Fbada006: Okay

checkout/src/main/java/com/payoneer/checkout/form/Operation.java
Fbada006: Ah, I see. That is handy and good to know. Thanks.

checkout/src/main/java/com/payoneer/checkout/risk/RiskException.java
Fbada006: I agree with you about the documentation @bzlat. However, in the rest of the code, @siebeprojects and I have not really been using a commenting strategy but rather writing code that explains itself without the need of comments. I am just trying to stick to a pattern that we have been following. This is ultimately your call @siebeprojects

checkout/src/main/java/com/payoneer/checkout/risk/RiskProviderController.java
Fbada006: Agreed.

checkout/src/main/java/com/payoneer/checkout/risk/RiskProviderLookup.java
Fbada006: Does the `Objects.require` give a better message when an NPE occurs? I am happy to refactor if it is better.

checkout/src/main/java/com/payoneer/checkout/risk/RiskException.java
siebeprojects: Also for inheriting docs, it is inconsistent in the Android SDK. 
We need to have a consistent approach and refactor this as well in another ticket.

checkout/src/main/java/com/payoneer/checkout/core/WorkerTask.java
siebeprojects: I updated this comment, but if we remove this one, we could remove all of them. I would like to leave this for another ticket where we refactor all our javadocs.

checkout/src/main/java/com/payoneer/checkout/risk/RiskProviderLookup.java
siebeprojects: No, they do not and therefor I have changed them to IllegalArgumentExceptions.

shared-test/lists/listresult_risk.json
siebeprojects: For now, this URL is used to test Iovation list results until we have a test merchant setup.

checkout/src/main/java/com/payoneer/checkout/core/WorkerTask.java
Fbada006: Okay

checkout/src/main/java/com/payoneer/checkout/risk/RiskException.java
Fbada006: That is all i ASK. Consistency

Fbada006 (approved):

PR #305, opened 2022-02-21
PCX-288: Remove toString statements containing sensitive data

siebeprojects:


checkout/src/main/java/com/payoneer/checkout/form/BrowserDataBuilder.java
```diff
@@ -89,30 +89,4 @@ public BrowserDataBuilder setBrowserScreenHeight(Integer browserScreenHeight) {
...
-
-    @Override
-    public String toString() {
-        final StringBuilder builder = new StringBuilder();
-        builder.append("BrowserDataBuilder [");
-        if (javaEnabled != null) {
-            builder.append("javaEnabled=").append(javaEnabled).append(", ");
-        }
-        if (language != null) {
-            builder.append("language=").append(language).append(", ");
```
Fbada006, line 101: Is this enough? Someone could still access these values if they have the object, right? My Java is failing me for a second here so bear with me

siebeprojects: It is ok that they can access the values, what I would like to avoid is that somebody would add a statement like Log.i("checkout", "Checkout values: " + browserData) and forgets to remove it and that would log all parameters.

Fbada006 (approved):

bzlat (approved):

PR #306, opened 2022-02-22
Feature/PCX-2615

siebeprojects:
**AS A** Payoneer risk manager

**I WANT** the mobile SDK to collect and submit Iovation risk-related data on the customer device and submit it with the operation calls (CHARGE, PRESET, etc.)

**SO THAT** we can use Iovation to assess each transaction risk and minimise our fraud-related liability

**Acceptance criteria**

- Given that the riskProviders object exists in the LIST object and contains a provider object with providerCode IOVATION1:
- The Iovation library is initialised with the parameters in the provider object
- The library initialisation happens at the earliest point in time possible without having a negative effect on UI rendering and user interaction
- The information collected and returned by the library’s collection method is submitted to OPG in operation calls of all types (e.g. CHARGE, PRESET, PAYOUT, etc.) in the structure defined in [ROP-431: [OPG CHARGE] Add new field "providerRequests"DEV DONE](https://optile.atlassian.net/browse/ROP-431)
- The collection process does not interrupt with the customer’s ability use the CTA button (e.g. click the “Pay”, “Continue” buttons, etc.)
- If the Iovation library is not present (was not compiled with the application) or the library’s collection method did not complete before the operation end-point was called, an “empty” (no parameters) risk provider object is submitted with operation


checkout/src/main/java/com/payoneer/checkout/risk/RiskProviderController.java
```diff
@@ -82,18 +82,19 @@ public void initialize(final Context context) {
      * @return RiskProviderResult obtained from the RiskProvider
      */
     public RiskProviderResult getRiskProviderResult(final Context context) {
+        RiskProviderResult result = null;
         if (riskProvider != null) {
             String code = info.getRiskProviderCode();
             String type = info.getRiskProviderType();
```
bzlat, line 88: nitpick: these two are only used in the `catch` block and could be moved there

checkout/src/main/java/com/payoneer/checkout/risk/RiskProviderController.java
```diff
@@ -82,18 +82,19 @@ public void initialize(final Context context) {
...
                 Context applicationContext = context.getApplicationContext();
-                return riskProvider.getRiskProviderResult(applicationContext);
+                result = riskProvider.getRiskProviderResult(applicationContext);
             } catch (RiskException e) {
                 String message = "RiskProviderController(" + code + ", " + type + ") could not obtain result";
                 Log.w("checkout-sdk", message, e);
             }
         }
-        return new RiskProviderResult();
+        return (result != null) ? result : new RiskProviderResult();
```
bzlat, line 98: nitpick: assign `result = new RiskProviderResult();` in the `catch` block, and just return the `result`. This conditional here is unnecessary

bzlat (approved):

Fbada006 (approved):

PR #312, opened 2022-03-04
PCX-2872: Move Iovation risk provider to subfolder riskproviders

siebeprojects:
**Description**
Move risk provider iovation in a subfolder called riskprovider and not in a root folder called riskprovider-iovation.

**Acceptance Criteria**
All risk providers are stored in the folder riskproviders:
for example:
/checkout
/riskprovider/iovation
/riskprovider/riskified


example-checkout/build.gradle
```diff
@@ -85,8 +85,8 @@ dependencies {
     implementation project(":checkout")
 
     // Comment the following two lines for disabling the Iovation FraudForce risk provider
-    implementation project(":riskprovider-iovation")
-    implementation files('../riskprovider-iovation/FraudForce/fraudforce-lib-release-4.3.2.aar')
+    // implementation project(":riskprovider:iovation")
```
Fbada006, line 88: What is happening here?

settings.gradle
```diff
@@ -1 +1,5 @@
-include ':checkout', ':example-checkout', ':example-shop', ':shared-test', 'riskprovider-iovation'
+include ':checkout',
+        ':example-checkout',
```
Fbada006, line 2: You have confirmed that if you add the module in the subfolder does not generate unnecessary empty modules?

example-checkout/build.gradle
siebeprojects: Ah, by default they should be commented because we should not have risk enabled by default, I will change the comment which should actually mention to "uncomment:.

Fbada006 (approved):

PR #317, opened 2022-03-15
PCX-2888: Remove caching per list for localization files

siebeprojects:
## WHAT

Cached localization files remain in memory cache until the application is closed.
Remove the cleanup of cache when the Checkout SDK is opened with a different list Url

## WHY

To improve the performance of using the Android Checkout SDK with different listUrls e.g. Update & Charge flow for the duration of that the app is running.

See https://optile.atlassian.net/browse/PCX-2888


checkout/src/main/java/com/payoneer/checkout/ui/service/PaymentSessionService.java
```diff
@@ -184,11 +184,6 @@ private void loadValidator(Context context) throws PaymentException {
     }
 
     private void loadLocalizations(final PaymentSession session, final Context context) throws PaymentException {
-        String listUrl = session.getListSelfUrl();
-        if (!Objects.equals(listUrl, cache.getCacheId())) {
```
Fbada006, line 188: Removing these lines makes the SDK faster @siebeprojects ?

Fbada006 (approved):

checkout/src/main/java/com/payoneer/checkout/ui/service/PaymentSessionService.java
siebeprojects: Currently, cached localization files are bound to the listUrl, if the merchant would reopen the payment list with a new listUrl, this would mean that all localization files are reloaded from the network. 

This is expensive and not needed, because localization files are not bound to a list.

PR #321, opened 2022-03-24
PCX-2974: Remove recurrence and use mastercard

siebeprojects:
## WHY

Setting the recurrence flag to true will not allow deletion of accounts. The merchant MOBILE_TESTING on  pi-nightly has changed the settings for VISA, whereby the recurrence flag should now be “false”.

See https://optile.atlassian.net/browse/PCX-2974

### DIFF
--- a/shared-test/src/main/java/com/payoneer/checkout/sharedtest/checkout/TestDataProvider.java
+++ b/shared-test/src/main/java/com/payoneer/checkout/sharedtest/checkout/TestDataProvider.java
@@ -27,6 +27,15 @@ public final class TestDataProvider {
         return values;
     }
 
+    public static Map<String, String> masterCardTestData() {
+        Map<String, String> values = new LinkedHashMap<>();
+        values.put("inputelement.number", "5555555555554444");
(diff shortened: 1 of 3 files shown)


shared-test/src/main/java/com/payoneer/checkout/sharedtest/checkout/TestDataProvider.java
```diff
@@ -27,6 +27,15 @@
         return values;
     }
 
+    public static Map<String, String> masterCardTestData() {
+        Map<String, String> values = new LinkedHashMap<>();
+        values.put("inputelement.number", "5555555555554444");
+        values.put("inputelement.expiryDate", "1245");
+        values.put("inputelement.verificationCode", "123");
+        values.put("inputelement.holderName", "Thomas Smith");
+        return values;
```
bzlat, line 36: Suggestion: Add test utils containing preconfigured credit card stubs and just use them everywhere, instead of repeating the same dummy numbers everywhere as `magic` numbers.

bzlat (approved):

Fbada006 (approved):

PR #325, opened 2022-04-14
PCX-2998: Report risk provider errors

Fbada006:
This PR adds parameters for internal and external errors during risk initialisation to the items sent back to OPG.


caiomello (dismissed):

checkout/src/main/java/com/payoneer/checkout/risk/RiskErrors.java
```diff
@@ -0,0 +1,46 @@
...
+ * This class is basically an extension of {@link com.payoneer.checkout.risk.RiskProviderResult} with a map
+ * of the external and internal errors to send back to OPG as part of the result object
+ */
+public class RiskErrors {
+
+    public static final String RESULTKEY_EXTERNAL_ERROR = "riskPluginExternalError";
+    public static final String RESULTKEY_INTERNAL_ERROR = "riskPluginInternalError";
+    private final List<Parameter> riskErrorParameters = new ArrayList<>();
+
+    public void addErrorParameter(Parameter parameter) {
```
siebeprojects, line 28: Instead of addErrorParameter() method, it would be better to define two methods:

- addInternalError(String errorMessage) 
- addExternalError(String errorMessage) 

These functions would create the Parameter object and store it into the list of parameters.

checkout/src/main/java/com/payoneer/checkout/risk/RiskProviderController.java
```diff
@@ -62,14 +68,22 @@ public void initialize(final Context context) {
 
         if (riskProvider == null) {
             String message = "RiskProviderController(" + code + ", " + type + ") could not find RiskProvider";
+            Parameter parameter = new Parameter();
```
siebeprojects, line 71: No need to create the parameter class here if calling the "addInternalError" method in the RiskErrors class.

checkout/src/main/java/com/payoneer/checkout/risk/RiskProviderController.java
```diff
@@ -62,14 +68,22 @@ public void initialize(final Context context) {
...
             Log.w("checkout-sdk", message);
             return;
         }
         try {
             Context applicationContext = context.getApplicationContext();
             riskProvider.initialize(info, applicationContext);
         } catch (RiskException e) {
-            String message = "RiskProviderController(" + code + ", " + type + ") failed to initialize RiskProvider";
+            String message = "RiskProviderController(" + code + ", " + type + ") failed to initialize RiskProvider " + e.getMessage();
+            Parameter parameter = new Parameter();
```
siebeprojects, line 83: No need to create the parameter class here if calling the "addExternalError" method in the RiskErrors class.

checkout/src/main/java/com/payoneer/checkout/risk/RiskProviderController.java
```diff
@@ -91,10 +105,18 @@ public RiskProviderResult getRiskProviderResult(final Context context) {
                 Context applicationContext = context.getApplicationContext();
                 result = riskProvider.getRiskProviderResult(applicationContext);
             } catch (RiskException e) {
-                String message = "RiskProviderController(" + code + ", " + type + ") could not obtain result";
+                String message = "RiskProviderController(" + code + ", " + type + ") could not obtain result " + e.getMessage();
+                Parameter parameter = new Parameter();
```
siebeprojects, line 109: No need to create the parameter class here if calling the "addExternalError" method in the RiskErrors class.

checkout/src/main/java/com/payoneer/checkout/risk/RiskProviders.java
```diff
@@ -124,8 +124,9 @@ private ProviderParameters getRiskProviderRequest(final RiskProviderController c
         List<Parameter> parameters = new ArrayList<>();
         request.setParameters(parameters);
 
+        RiskErrors errors = controller.getRiskErrors();
         RiskProviderResult result = controller.getRiskProviderResult(context);
-        result.copyInto(parameters);
+        result.copyInto(parameters, errors.getRiskErrorParameters());
```
siebeprojects, line 129: Both do not need to be add both lists in one method.

- If there are errors then we only add error parameters
- If there are no errors then add the risk data parameters

Perhaps a method called "RiskErrors.isEmpty()" or RiskErrors.hasErrors()" could be used to determine which list (risk errors/risk data) to add to the RiskProviderResult class.

checkout/src/main/java/com/payoneer/checkout/risk/RiskErrors.java
Fbada006: Done

checkout/src/main/java/com/payoneer/checkout/risk/RiskProviderController.java
Fbada006: Done

checkout/src/main/java/com/payoneer/checkout/risk/RiskProviderController.java
Fbada006: Done

checkout/src/main/java/com/payoneer/checkout/risk/RiskProviders.java
Fbada006: Agreed. I have made some changes. See how they are inside the definition of the `copyInto` method.

siebeprojects (approved):

checkout/src/main/java/com/payoneer/checkout/risk/RiskProviderErrors.java
```diff
@@ -0,0 +1,67 @@
...
+
+package com.payoneer.checkout.risk;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * This class is used to store internal and external errors generated by a risk provider.
+ */
+public final class RiskProviderErrors {
```
caiomello, line 19: Should this be plural? Why not use a collection/array of errors?

caiomello (approved):

checkout/src/main/java/com/payoneer/checkout/risk/RiskProviderErrors.java
siebeprojects: It is a storage for multiple risk provider errors, it cannot be singular. 
Inside it has a map to store the errors.

PR #344, opened 2022-05-24
PCX-3317: Disable reload when payment list is resumed

siebeprojects:
## Jira ticket
[PCX-3317](https://optile.atlassian.net/browse/PCX-3317)

## Overview/What
When clicking an ExtraElement link which opens the custom chrome tab. After closing the custom tab, the app is reopened but the list is reloaded.

## Cause/Why
The Mobile SDK does not check if there is a previously loaded payment session. It always loads a new one when the PaymentList Screen is resumed.

## Solution
Show the already loaded payment session, if previously loaded.

## Type of change
- Bug fix (non-breaking change which fixes an issue)
- Test cases

## Tests
- Automated test has been added to validate that the correct payment card is opened when returning from the Chrome Custom tab.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -95,13 +95,13 @@ ext {
 }
 
 static def getBranchName() {
-    def gitRefName = System.getenv('GITHUB_REF_NAME')
-    if (gitRefName != null) {
-        return gitRefName
-    }
     def gitHeadRef = System.getenv('GITHUB_HEAD_REF')
     if (gitHeadRef != null) {
-        return gitHeadRef
+        return "HR" + gitHeadRef
+    }
+    def gitRefName = System.getenv('GITHUB_REF_NAME')
+    if (gitRefName != null) {
+        return "RN" + gitRefName
     }
     def branch = ''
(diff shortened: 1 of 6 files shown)


caiomello (approved):

caiomello (dismissed):

build.gradle
```diff
@@ -95,6 +95,14 @@ ext {
...
 
 static def getBranchName() {
+    def gitHeadRef = System.getenv('GITHUB_HEAD_REF')
+    if (gitHeadRef != null) {
+        return gitHeadRef
+    }
+    def gitRefName = System.getenv('GITHUB_REF_NAME')
+    if (gitRefName != null) {
+        return gitRefName
+    }
```
Sorix, line 105: Does it have a relation to this ticket?

Fbada006 (dismissed):

build.gradle
siebeprojects: Removed from this ticket and moved to PCX-3340.

Sorix (approved):

Fbada006 (approved):

PR #345, opened 2022-05-25
PCX-3313: Delete AppLive browserstack apps with customId

siebeprojects:
## Jira ticket
[PCX-3313](https://optile.atlassian.net/browse/PCX-3313)

## Overview/What
When uploading a new app version to BrowserStack AppLive for a branch, any previously uploaded app versions for that branch should be deleted first. 

## Cause/Why
As an integration engineer / PO / Support person
I like to only see one instance of a specific app for a branch in AppLive
so that it is easy to know which app is the latest version

## Solution
Upload apps to Browserstack with a customId.  
The customId is constructed from the name of the app and branch name.

## Type of change


browserstack.gradle
```diff
@@ -0,0 +1,29 @@
...
+        return
+    }
+    for (def app : (List<Map>) json) {
+        if (!(app.containsKey("app_id"))) {
+            continue
+        }
+        logger.lifecycle("[BROWSERSTACK] Deleting app with customId[${customId}] and appId[${app.app_id}]")
+        exec {
+            executable "curl"; args "-u", "$rootProject.browserStackUser:$rootProject.browserStackKey", "-X",
+                    "DELETE", "https://api-cloud.browserstack.com/app-live/app/delete/${app.app_id}"
```
Sorix, line 26: If `app_id` will contain `1; injected-command` will it be executed or it is safe to pass this argument?

build.gradle
```diff
@@ -107,4 +109,4 @@ static def getBranchLabel() {
     return getBranchName().replace('/', '-')
```
Sorix, line 109: A branch name could contain other illegal characters, I recommend replacing all non-alphanumeric characters (excluding dashes`-`).

browserstack.gradle
```diff
@@ -0,0 +1,29 @@
...
+    def response = new ByteArrayOutputStream()
+    exec {
+        executable "curl";
+        args "-u", "$rootProject.browserStackUser:$rootProject.browserStackKey", "-X",
+                "GET", "https://api-cloud.browserstack.com/app-live/recent_apps/${customId}"
+        standardOutput = response;
+    }
+    def json = new JsonSlurper().parseText(response.toString())
+    if (!(json instanceof List<Map>)) {
+        logger.lifecycle("[BROWSERSTACK] No apps to delete for customId[${customId}]")
```
Sorix, line 16: If we don't parse error messages we should mention this in a log.

```suggestion
        logger.lifecycle("[BROWSERSTACK] No apps to delete for customId[${customId}] or error response was received from a server")
```

browserstack.gradle
siebeprojects: Add urlEncoding for the appId.

browserstack.gradle
siebeprojects: Added the json response to the log message.

Sorix (approved):

PR #359, opened 2022-06-13
PCX-3458: Add unit tests in the Checkout SDK

Fbada006:
## Jira ticket
[PCX-3458](https://optile.atlassian.net/browse/PCX-3458)

## Overview/What
Add viewmodel tests to the SDK

## Cause/Why
To get more test coverage in line with company goals

## Type of change
- Test cases

## Tests
- ProcessPaymentViewModelTest
- PaymentListViewModelTest


siebeprojects (dismissed):

checkout/src/test/java/com/payoneer/checkout/payment/PaymentServiceViewModelTest.java
```diff
@@ -0,0 +1,17 @@
...
+ *
+ */
+
+package com.payoneer.checkout.payment;
+
+import static org.junit.Assert.*;
+
+public class PaymentServiceViewModelTest {
+
+}
```
Sorix, line 17: What's the purpose of the empty class?

checkout/src/test/java/com/payoneer/checkout/ui/screen/payment/ProcessPaymentViewModelTest.java
```diff
@@ -0,0 +1,58 @@
...
+        ProgressSettings progressSettings = LiveDataUtil.getOrAwaitValue(viewModel.showProcessPaymentProgress()).getContentIfNotHandled();
+
+        assertNotNull(event);
+        assertNotNull(progressSettings);
+        assertTrue(progressSettings.visible);
+    }
+
+    private URL createUrl() {
+        try {
+            return new URL("https://raw.githubusercontent.com/optile/checkout-android/develop/shared-test/lists/listresult.json");
```
Sorix, line 53: Why do we still use list result responses stored in such a way? I thought we decided not to use it and we don't use it in iOS.

checkout/src/test/java/com/payoneer/checkout/ui/screen/payment/ProcessPaymentViewModelTest.java
```diff
@@ -0,0 +1,58 @@
...
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
Sorix, line 55: I'm not sure how it is done in Java, but should you catch and throw the same error? What's the goal of catch in that case?

checkout/src/test/java/com/payoneer/checkout/util/LiveDataUtil.java
```diff
@@ -0,0 +1,44 @@
...
+        Observer<T> observer1 = new Observer<T>() {
+            @Override
+            public void onChanged(T t) {
+                data[0] = t;
+                latch.countDown();
+                liveData.removeObserver(this);
+            }
+        };
+
+        liveData.observeForever(observer1);
```
Sorix, line 35:
```suggestion
        Observer<T> observer = new Observer<T>() {
            @Override
            public void onChanged(T t) {
                data[0] = t;
                latch.countDown();
                liveData.removeObserver(this);
            }
        };

        liveData.observeForever(observer);
```

checkout/src/test/java/com/payoneer/checkout/payment/PaymentServiceViewModelTest.java
Fbada006: This should not be here. I thought I removed this. I am deleting it

checkout/src/test/java/com/payoneer/checkout/ui/screen/payment/ProcessPaymentViewModelTest.java
Fbada006: This is not a list result response. This is just creating a URL object

checkout/src/test/java/com/payoneer/checkout/ui/screen/payment/ProcessPaymentViewModelTest.java
Fbada006: This is a mistake, which I have corrected.

checkout/src/test/java/com/payoneer/checkout/util/LiveDataUtil.java
```diff
@@ -0,0 +1,44 @@
...
+
+        // Don't wait indefinitely if the LiveData is not set.
+        if (!latch.await(2, TimeUnit.SECONDS)) {
```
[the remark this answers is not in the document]
Fbada006: Alright.

checkout/src/test/java/com/payoneer/checkout/ui/screen/list/PaymentListViewModelTest.java
```diff
@@ -0,0 +1,161 @@
...
+        CheckoutResult result = error.getContentIfNotHandled();
+        Interaction interaction = result.getInteraction();
+        assertNotNull(result);
+        assertEquals(interaction.getCode(), "ABORT");
+        assertEquals(interaction.getReason(), "CLIENTSIDE_ERROR");
+    }
+
+    private URL createUrl() {
+        try {
+            return new URL("https://raw.githubusercontent.com/optile/checkout-android/develop/shared-test/lists/listresult.json");
```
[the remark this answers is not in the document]
Fbada006: Again, this is just the creation of a url object, not the list result response

checkout/src/test/java/com/payoneer/checkout/ui/screen/payment/ProcessPaymentViewModelTest.java
Sorix: Will the new code lead to `NullPointerException`? Isn't, in that case, it's better not to catch it?

checkout/src/test/java/com/payoneer/checkout/ui/screen/payment/ProcessPaymentViewModelTest.java
Fbada006: Alright

checkout/src/test/java/com/payoneer/checkout/ui/screen/payment/ProcessPaymentViewModelTest.java
Fbada006: The new code just catches it and prints the stack trace

siebeprojects: It would be better to catch the MalformedURLException instead. Its better to catch the specific Exception than the generic one.

checkout/src/test/java/com/payoneer/checkout/ui/screen/payment/ProcessPaymentViewModelTest.java
Fbada006: This has been changed

checkout/src/test/java/com/payoneer/checkout/ui/screen/payment/ProcessPaymentViewModelTest.java
Sorix: Let me try to clarify (at least what I understand from this code). You create a URL with `NULL` value. You try to assign a value or print a stack trace. What will happen if the URL couldn't be built? As I understand value will be `NULL` and it will be used as the return value of a method, so you return `NULL` instead of expected `URL`.

https://github.com/optile/checkout-android/blob/0694ac79eccb8e0635b709bf54d6f6ae88c92395/checkout/src/test/java/com/payoneer/checkout/ui/screen/payment/ProcessPaymentViewModelTest.java#L52-L60

siebeprojects: Thats correct, when the test needs or expects a non-null URL, its better to have a statement in the test like:
```
URL url = createUrl();
assertNotNull(url);

```
This MalformedURLException will never be thrown but it is still better to check in the test method itself that the url is not null.

siebeprojects: Or the test method could be like:

```
@Test
public void testFoo() throws MalformedURLException {
    URL url = createUrl();
```

siebeprojects (approved):

Sorix (approved):

PR #367, opened 2022-07-11
PCX-3630 Switched to a proper value for the providerCode when finaliz…

izpakla:
…ing payment via GooglePay

## Jira ticket
[PCX-3630](https://optile.atlassian.net/browse/PCX-3630)

## Overview/What
Making a GooglePay charge request using direct charge or preset flow, it fails with a TRY_OTHER_ACCOUNT/INVALID_ACCOUNT interaction code and reason.

## Cause/Why
The GooglePayBraintreePaymentService sends in the providerRequest the network code for the providerCode value. This is incorrect.

## Solution
The GooglePayBraintreePaymentService should send the providerCode value, received in the onSelect result, for the final charge request.

## Type of change
- Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/paymentservice/googlepay-braintree/src/main/java/com/payoneer/checkout/payment/googlepaybraintree/GooglePayBraintreePaymentService.java
+++ b/paymentservice/googlepay-braintree/src/main/java/com/payoneer/checkout/payment/googlepaybraintree/GooglePayBraintreePaymentService.java
@@ -94,6 +94,7 @@ public class GooglePayBraintreePaymentService extends PaymentService {
         this.processPaymentData = null;
         this.applicationContext = null;
         this.fragmentResult = null;
+        this.providerCode = null;
         operationService.stop();
     }
 
@@ -208,10 +209,12 @@ public class GooglePayBraintreePaymentService extends PaymentService {
             return;
         }
 
-        this.providerCode = PaymentUtils.getProviderCode(operationResult);
-        if (TextUtils.isEmpty(this.providerCode)) {
+        String providerCode = PaymentUtils.getProviderCode(operationResult);
+        if (TextUtils.isEmpty(providerCode)) {
             closeWithProcessErrorMessage("Missing GooglePayBraintree provider code");
             return;
+        } else {
+            this.providerCode = providerCode;
         }
 
         try {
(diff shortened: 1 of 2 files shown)


paymentservice/googlepay-braintree/src/main/java/com/payoneer/checkout/payment/googlepaybraintree/GooglePayBraintreePaymentService.java
```diff
@@ -137,14 +138,18 @@ private void handleGetNonceResult() {
             closeWithProcessErrorMessage("Missing GooglePayBraintree fragment result after onResume");
             return;
         }
+        if (providerCode == null) {
```
Fbada006, line 141: Why not use TextUtils.isEmpty here just like in the other check?

izpakla: Since other check sets it, there are only two options here, null or correct value. That's why I only left null check.

Fbada006: Fair enough

Fbada006 (approved):

paymentservice/googlepay-braintree/src/main/java/com/payoneer/checkout/payment/googlepaybraintree/GooglePayBraintreePaymentService.java
siebeprojects: @izpakla do not forget to reset the providerCode to null in the reset() method.

In theory, because you do a TExtUtils.isEmpty() when setting the providerCode, it could be an empty string.
When you do not reset this value when we bail out from the payment because of an error, the providerCode could remain being an empty string and then this check would fail.

paymentservice/googlepay-braintree/src/main/java/com/payoneer/checkout/payment/googlepaybraintree/GooglePayBraintreePaymentService.java
```diff
@@ -202,6 +207,13 @@ private void handleProcessOnSelectSuccess(final OperationResult operationResult)
             closeWithProcessErrorMessage("Missing GooglePayBraintree [" + BRAINTREE_AUTHORIZATION + "] parameter");
             return;
         }
+
+        this.providerCode = PaymentUtils.getProviderCode(operationResult);
```
siebeprojects, line 211: _Not a blocker_
Normally, it is better to set the global class member after the value is validated, because setting the global class member here allows it to become an empty string while that is not desired.
Perhaps, setting it first in a local variable, then validate the local variable and when all is fine, set the global class member.

But, when we reset the providerCode to null in the reset() method, then this will work as it is.

siebeprojects (approved):

PR #368, opened 2022-07-19
PCX-3617 Extended service factory's supports method with providers li…

izpakla:
…st. Switched Braintree's condition to providers list.

## Jira ticket
[PCX-3617](https://optile.atlassian.net/browse/PCX-3617)

## Overview/What
Lookup of PaymentServices in Android SDK using optional providerCode

## Solution
Extended Payment service validation with a list of providerCodes.

## Type of change

### DIFF
--- a/checkout/src/main/java/com/payoneer/checkout/payment/PaymentServiceFactory.java
+++ b/checkout/src/main/java/com/payoneer/checkout/payment/PaymentServiceFactory.java
@@ -8,6 +8,10 @@
 
 package com.payoneer.checkout.payment;
 
+import java.util.List;
+
+import androidx.annotation.Nullable;
+
 /**
  * Interface for all payment service factories.
  * A payment service factory is capable of creating a PaymentService instance for a specific PaymentNetwork type.
@@ -19,9 +23,10 @@ public interface PaymentServiceFactory {
      *
      * @param code to be checked if it is supported by this factory
      * @param method to be checked if it is supported by this factory
+     * @param providers to be checked if it is supported by this factory
      * @return true when supported, false otherwise
      */
-    boolean supports(String code, String method);
+    boolean supports(String code, String method, @Nullable List<String> providers);
 
     /**
      * Create a payment service for this specific payment network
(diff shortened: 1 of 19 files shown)


checkout/src/main/java/com/payoneer/checkout/payment/PaymentServiceFactory.java
```diff
@@ -19,9 +23,10 @@ public interface PaymentServiceFactory {
      *
      * @param code to be checked if it is supported by this factory
      * @param method to be checked if it is supported by this factory
+     * @param providers to be checked if it is supported by this factory
      * @return true when supported, false otherwise
      */
-    boolean supports(String code, String method);
+    boolean supports(String code, String method, @Nullable List<String> providers);
```
Fbada006, line 29: @siebeprojects , can we finally discuss these annotations in code? i hope you can attend the meeting by @izpakla tomorrow so we finalise this

checkout/src/main/java/com/payoneer/checkout/payment/basic/BasicPaymentServiceFactory.java
```diff
@@ -22,7 +26,7 @@ public final class BasicPaymentServiceFactory implements PaymentServiceFactory {
      * {@inheritDoc}
      */
     @Override
-    public boolean supports(String code, String method) {
+    public boolean supports(String code, String method, @Nullable final List<String> providers) {
         switch (method) {
```
siebeprojects, line 30: There should be a check that if the providers list is non-null and has elements that this supports method must return false. Because this PaymentService supports payment methods from different providers and not for one specific.

izpakla: Added check.

checkout/src/main/java/com/payoneer/checkout/payment/PaymentServiceFactory.java
izpakla: Updated annotations and fixed null values passed in tests.

caiomello (approved):

PR #369, opened 2022-07-19
PCX-3503: Validate checkbox states

Fbada006:
## Jira ticket
[PCX-3503](https://optile.atlassian.net/browse/PCX-3503)

## Overview/What
Validate the checkbox modes before submitting them back to OPG. Specifically, validate the ones with either `REQUIRED` or `REQUIRED_PRESELECTED` modes

## Cause/Why
We want to make sure that the users do not send requests if they have not fully agreed to the merchant's conditions

## Solution
Add validation before the states are submitted to OPG

## Type of change

## Tests
- Changes have been tested in UI tests where we check that the error messages are displayed upon clicking the pay button and upon toggling the `REQUIRED` or `REQUIRED_PRESELECTED` check boxes.

### DIFF
--- a/checkout/src/main/java/com/payoneer/checkout/ui/widget/CheckboxWidget.java
+++ b/checkout/src/main/java/com/payoneer/checkout/ui/widget/CheckboxWidget.java
@@ -116,13 +116,9 @@ public class CheckboxWidget extends FormWidget {
             case CheckboxMode.OPTIONAL_PRESELECTED:
             case CheckboxMode.REQUIRED_PRESELECTED:
             case CheckboxMode.FORCED_DISPLAYED:
+            case CheckboxMode.FORCED:
                 setVisible(true);
                 switchView.setVisibility(View.VISIBLE);
-                switchView.setChecked(true);
-                break;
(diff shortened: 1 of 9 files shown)


izpakla (approved):

checkout/src/main/java/com/payoneer/checkout/ui/widget/CheckboxWidget.java
```diff
@@ -35,16 +46,52 @@ public CheckboxWidget(String category, String name) {
...
     @Override
     public void putValue(final PaymentInputValues inputValues) {
         inputValues.putBooleanValue(category, name, switchView.isChecked());
     }
 
+    @Override
+    public boolean validate() {
+        boolean checked = switchView.isChecked();
+        switch (mode) {
+            case REQUIRED:
```
Sorix, line 72: Will it fallthrough and make validation in `REQUIRED_PRESELECTED` (so the validation message will be shown if it was `REQUIRED` but not selected)?

Sorix (approved):

PR #372, opened 2022-07-26
PCX-3506: show information why extra element cannot be unchecked

Fbada006:
## Jira ticket
[PCX-3506](https://optile.atlassian.net/browse/PCX-3506)

## Overview/What
Show a dialog when a user tries to deselect a checkbox with the mode `FORCED` or `FORCED_DISPLAYED`

## Cause/Why
So that the end customers can know why they cannot uncheck a checkbox

## Solution
Set a listener to the checkbox in the generic `CheckBoxWidget` class then make sure the switch is always checked for the above two modes. Then show a dialog with the message

## Type of change

## Tests
- Testing that the dialog has been displayed when the two checkboxes with the above modes are clicked
- Happy path when all is selected then we get a success from the backend


checkout/src/main/java/com/payoneer/checkout/ui/widget/CheckboxWidget.java
```diff
@@ -57,7 +59,8 @@ public View inflate(ViewGroup parent) {
     }
 
     private void handleSwitchClicked(final boolean isChecked) {
-        showRequiredMessage(!isChecked && (Objects.equals(mode, REQUIRED) || Objects.equals(mode, REQUIRED_PRESELECTED)));
+        showRequiredMessage(!isChecked && (Objects.equals(mode, CheckboxMode.REQUIRED) || Objects.equals(mode, CheckboxMode.REQUIRED_PRESELECTED)));
+        handleForcedCheckBoxClick(isChecked);
     }
```
Sorix, line 65: I'm not sure what's going on in this method, could you describe it?

Fbada006: We are listening to clicks on the switch so that `isChecked` argument keeps on changing as the user clicks. Then we check the mode that it is `FORCED` or `FORCED_DISPLAYED` and we se the switch to checked immediately if  `isChecked` is false.
Is this clear @Sorix ?

Sorix: `showRequiredMessage(Objects.equals(mode, CheckboxMode.REQUIRED_PRESELECTED)))`

This will show the required message even if the switch was checked?

Fbada006: The required message is only displayed for the required and required_preselected modes. 
And yes, this method ensures the required message is displayed for the above modes.

Fbada006: Also, the actual method is active for either required and required preselected. What you've pasted is not quite right.
Not the || operator, which is an OR operator

siebeprojects: @Fbada006 it would be better to implement here a similar switch statement as done in the validate method.
Then it is not needed to call always both methods but only call the ones that are needed based on the current mode.

checkout/src/main/java/com/payoneer/checkout/ui/widget/CheckboxWidget.java
```diff
@@ -92,6 +95,15 @@ private void showRequiredMessage(final boolean show) {
         }
     }
 
+    private void handleForcedCheckBoxClick(final boolean isChecked) {
+        if ((Objects.equals(mode, CheckboxMode.FORCED) || Objects.equals(mode, CheckboxMode.FORCED_DISPLAYED))) {
```
siebeprojects, line 99: When there is a switch statement in the handleSwitchClicked method then we do not need this if statement.

checkout/src/main/java/com/payoneer/checkout/ui/widget/CheckboxWidget.java
Sorix: I don’t block, but in my opinion, this code doesn't have good readability, at least for non-Java developers. Using `||`, `&&` and `!/not` at the same call makes too complex logic in one line.

If you have to use `&&` and `||` at the same line, I think logic should be separated by brackets like `((a && b) || c))`.

But I’m against putting extra logic in one line, I vote for having multiple lines but better readability.

checkout/src/main/java/com/payoneer/checkout/ui/widget/CheckboxWidget.java
Fbada006: Alright.
I can do that

checkout/src/main/java/com/payoneer/checkout/ui/widget/CheckboxWidget.java
Fbada006: Thanks team. Changes made

Sorix (approved):

siebeprojects (approved):

siebeprojects (approved):

PR #380, opened 2022-08-17
PCX-3784: display key name when required message is empty

Fbada006:
## Jira ticket
[PCX-3784](https://optile.atlassian.net/browse/PCX-3784)

## Overview/What
Display key name when required message is empty

## Cause/Why
The Backend Merchant Configuration API is not supposed to allow an empty requiredMessage for an extraElement with a checkbox. This is the validation message that should appear in case the end-customer did not check a REQUIRED or a REQUIRED_PRESELECT checkbox.)

## Solution
If requiredMessage for an extraElement with a checkbox is somehow missing nevertheless, and if a requiredMessage needs to be shown to the end-customer since a required checkbox is switched off (when pre-selected) or a submission attempt is made without the checkbox checked (when not pre-selected), then the Android SDK displays a (non-localised) string as the validation message in the format “<extraElementName>.requiredMessage”

## Type of change

## Tests
- Unit tests for the required message functionality in the PaymentUtilsTest class


checkout/src/main/java/com/payoneer/checkout/util/PaymentUtils.java
```diff
@@ -401,4 +405,25 @@ public static <K, V> Map<K, V> emptyMapIfNull(Map<K, V> map) {
...
+     * @return the required message for a checkbox
+     */
+    public static String getCheckboxRequiredMessage(final ExtraElement extraElement) {
+        Checkbox checkbox = extraElement.getCheckbox();
+        if (checkbox == null) {
+            return null;
+        }
+        switch (checkbox.getMode()) {
+            case CheckboxMode.REQUIRED:
+            case CheckboxMode.REQUIRED_PRESELECTED:
```
caiomello, line 422: Is this the same as this?

```swift
case CheckboxMode.REQUIRED, CheckboxMode.REQUIRED_PRESELECTED:
```

checkout/src/main/java/com/payoneer/checkout/util/PaymentUtils.java
```diff
@@ -401,4 +405,25 @@ public static <K, V> Map<K, V> emptyMapIfNull(Map<K, V> map) {
...
+                String reqMessage = checkbox.getRequiredMessage();
+                return TextUtils.isEmpty(reqMessage) ? extraElement.getName() + ".requiredMessage" : reqMessage;
```
caiomello, line 424: I try to be careful with ternary operators since it's easy to make it overly complex and hard to read. In this situation I'd do something like this:

```suggestion
                String requiredMessage = checkbox.getRequiredMessage();
                String isRequiredMessageEmpty = TextUtils.isEmpty(requiredMessage)
                String fallbackMessage = extraElement.getName() + ".requiredMessage"
                return isRequiredMessageEmpty ? fallbackMessage : requiredMessage;
```

👍, 👎, thoughts?

caiomello (approved):

checkout/src/main/java/com/payoneer/checkout/util/PaymentUtils.java
Fbada006: No. You cannot do that with a switch in java sadly

checkout/src/main/java/com/payoneer/checkout/util/PaymentUtils.java
Fbada006: Changing this now

checkout/src/main/java/com/payoneer/checkout/util/PaymentUtils.java
caiomello: But this case doesn't return anything, is there an implicit `return null;` there?

caiomello (approved):

checkout/src/main/java/com/payoneer/checkout/util/PaymentUtils.java
Fbada006: The way it is now, the return statement is handling both cases namely `REQUIRED` and `REQUIRED_PRESELECTED`. The way to read this is if the case is `REQUIRED` or `REQUIRED_PRESELECTED`, then do the computation for the required message. If the two cases were handled differently, I would add the `break` keyword at the end of each case like below:

`
 switch (mode) {
            case CheckboxMode.OPTIONAL:
            case REQUIRED:
              // Handle cases for both required and optional above
                break;
            case CheckboxMode.OPTIONAL_PRESELECTED:
              // Handle just optional preselected and break out
                break;
            default:
              // This works just like an else statement
        }
`

Doe this make sense?

caiomello: That's what I meant by this:

```swift
case CheckboxMode.REQUIRED, CheckboxMode.REQUIRED_PRESELECTED:
```

This means case `REQUIRED` **OR** case `REQUIRED_PRESELECTED`.

siebeprojects (approved):

PR #385, opened 2022-08-19
PCX-3840: Add unit tests for com.payoneer.checkout.util.event

Fbada006:
## Jira ticket
[PCX-3840](https://optile.atlassian.net/browse/PCX-3840)

## Overview/What
Add tests in Event class

## Cause/Why
To drive coverage and integrity of the SDK up

## Type of change
Test cases


caiomello (commented):
It doesn't look like any tests are being _added_ here, is it just the additional assert call?

Fbada006:
Hey @caiomello , yes. Only the assertion was needed to cover that class 100%. It already had tests but they were not 100

caiomello (approved):

siebeprojects (approved):

